PR #84, opened 2015-01-26
[v1] Rename module to o-forms

kaelig:


kaelig:
Note that this PR is merging into the branch `v1` which will serve as a meta-issue for a bigger PR.

PR #88, opened 2015-02-06
[V1] de-extendify

kaelig:
- No more `@extend`
- No more fieldsets: (use `div`s instead, since `fieldset` has cross-browser inconsistencies that are impossible to override)
- Removes `o-ft-forms__section` (which was used to reset `fieldset` elements)
- Renames all classes into `o-forms`
- BEM-ification of classes
- A few base mixins only (to be determined with the membership team)
- README update
- Demos updates (and removes deprecated ones)

### New field class names

```
.o-forms-text
.o-forms-textarea
.o-forms-checkbox
.o-forms-radio
.o-forms-select
```

Grepping the codebase for `o-forms-text` would return lines containing `o-forms-textarea`, is that going to be a problem and if so, should we name `o-forms-text` differently?


demos/checkboxes.html
```diff
@@ -7,55 +7,55 @@
 	<meta http-equiv="X-UA-Compatible" content="IE=Edge" />
 	<title>o-forms: checkboxes demo</title>
 	<meta name="viewport" content="initial-scale=1.0, width=device-width" />
-	<script src="//polyfill.webservices.ft.com/v1/polyfill.min.js"></script>
+	<script src="//polyfill.webservices.ft.com/v1/polyfill.min.js?features=modernizr:borderradius,default"></script>
```
AlbertoElias: I don't think having borderradius there is useful

src/scss/fields.scss
```diff
@@ -1,98 +1,63 @@
-@include oFormsClass(o-forms__field-group) {
+@if not $o-forms-is-silent {
+.o-forms-group {
```
AlbertoElias: If your adding the properties directly to the classes when not in silent mode, how are they used outside of silent mode?

kaelig: For now, not all the properties are available in silent mode.

Considering adding them on a per-class basis as Membership asks for them.

This one is a good candidate though!

AlbertoElias: Only partially supporting silent mode is kind of weird. I'm not very sure that's a good idea. @triblondon What do you think?

kaelig: This is an old PR, and actually this has been mixinified, see #89

AlbertoElias: I chose this class randomly, I think it's still interesting to discuss if partial silent mode support is a good idea for Origami.

Issue #81, opened 2014-12-31
Using o-ft-forms in silent mode


jamesnicholls:
I'm currently trying to use forms in silent mode. I understand silent mode is meant to require a little more work than non-silent mode but I think it's currently too cumbersome. For example, to use all radio button styles, you need to extend the base field placeholder, radio button placeholder, radio button label placeholder, radio button border placeholder and placeholders for the various states a radio button can be in.

It would be nice if you could simply include the features that you need (e.g. text inputs, selects, radio buttons) and receive all relevant styles, while excluding features that you don't (e.g. textareas, affixes).

jamesnicholls:
I would like to see this addressed for milestone 1.0.0 as I assume it would require a breaking change.

kaelig:
Thanks for the feedback, makes total sense. I've assigned it to the 1.0.0 milestone. (another reason for which this module needs a full rewrite!)

kaelig:
We have discussed this and the v1 makes it easier to specify what you want to include with helper mixins.

PR #89, opened 2015-02-06
[V1] Add mixins to reproduce styles in silent mode

kaelig:
This should solve some of the issues raised in #81

/ping @Thurst-ft @jamesnicholls


kaelig:
What do you guys think of this mixin:

```scss
@mixin oFormsCheckboxRadio($selectors: (radio: '.o-forms-radio', checkbox: '.o-forms-checkbox', label: '.o-forms-label')) {
```

I think it should make styling of radio and checkboxes easier in silent mode. If you could try it, it'd be great.

PR #93, opened 2015-02-13
Disabled states - fixes #73

### LINKED ISSUE #73: Disabled fields look a lot like buttons
Fields look like this when disabled:

screenshot 2014-12-04 18 50 42

Should we have a look at making them look more 'disabled' and less emphasized on?

kaelig:
- New disabled states for checkboxes and radio buttons (they deserve it, too)
- Disabled inputs don't look like buttons anymore (Fixes #73)

### Before

screenshot 2014-12-04 18 50 42

### After
screen shot 2015-02-13 at 15 00 50

### Colors
Text: default field background (30%) + default text (70%)
Border: default field background (50%) + default border (50%)
Background: default field background (15%) + default border (85%)

**Q: Why not RGBA (transparent colors)?**
A: because plain colours are compatible with older browsers.

**Q: Why mixing colours instead of creating new use cases?**
A: to avoid having to override lots of use cases in their code if a product wants to change the color scheme

#### Next to a normal field, for contrast:
screen shot 2015-02-13 at 15 00 59
screen shot 2015-02-13 at 15 00 50

#### Checkboxes

Opacity simply set to 40% because it's simpler than changing colours of checked/unchecked states in combination with the disabled state.

screen shot 2015-02-13 at 15 25 11

screen shot 2015-02-13 at 15 24 41

### DIFF
--- a/demos/src/radios.mustache
+++ b/demos/src/radios.mustache
@@ -10,7 +10,7 @@
 	</div>
 </div>
 <div>
-	<div class="o-forms-group o-forms--valid">
+	<div class="o-forms-group">
 		<label class="o-forms-label">Label (disabled)</label>
 		<input type="radio" name="radio2" value="1" class="o-forms-radio" id="radio21" checked="checked" disabled="disabled" />
 		<label for="radio21" class="o-forms-label">Radio 1</label>
--- a/demos/src/radios.mustache
+++ b/demos/src/radios.mustache
@@ -11,12 +11,12 @@
 </div>
 <div>
 	<div class="o-forms-group o-forms--valid">
-		<label class="o-forms-label">Label (valid)</label>
-		<input type="radio" name="radio2" value="1" class="o-forms-radio" id="radio21" checked="checked" />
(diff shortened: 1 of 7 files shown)


demos/src/radios.mustache
```diff
@@ -11,12 +11,12 @@
 </div>
 <div>
 	<div class="o-forms-group o-forms--valid">
-		<label class="o-forms-label">Label (valid)</label>
```
kaelig: Note for review: valid states were exactly the same as… default states. So no need for them

demos/src/radios.mustache
```diff
@@ -11,12 +11,12 @@
 </div>
 <div>
 	<div class="o-forms-group o-forms--valid">
```
AlbertoElias: Is the class o-forms--valid needed?

AlbertoElias:
Looks good

demos/src/radios.mustache
kaelig: Good catch, cheers

PR #95, opened 2015-03-18
Update silent mode variable name

kaelig:


main.scss
```diff
@@ -20,3 +20,7 @@
 @import "src/scss/field-errors";
 @import "src/scss/field-valid";
 @import "src/scss/global-and-section-messages";
+
+// Set silent mode back on again to avoid it being output twice
+$o-forms-is-silent: true;
+$o-ft-forms-is-silent: true;
```
triblondon: This is contra spec: http://origami.ft.com/docs/syntax/scss/#sass-variables.  The reason for the rule was because setting the values of config vars in a module creates potentially unpredictable, load-order-specific behaviour.

kaelig: Oh? How do we make sure modules aren't output more than once?

I've been using this technique in other modules (such as the grid https://github.com/Financial-Times/o-grid/blob/master/main.scss#L55), and the spec doesn't suggest a way to avoid multiple outputs.

triblondon: Origami policy is to use CSS optimiser to remove code duplication after the sass is built. We should stick to that policy until it changes (I don't object to having a discussion about changing the policy, but consistency and compliance trumps experimentation)

kaelig: I completely agree on the principle but unfortunately CSS optimisers don't de-dupe selectors that appear multiple times in the code, in order to preserve the cascade. 

As long as Sass has no "include once" directive, I'm not sure how we can do it another way than overriding the silent mode right after the module was outputted.

triblondon: Sure, so lets have that discussion as a group and if we make a decision, we can write it in the spec and everyone will then know to code that way.  In the meantime, I think we should stick to the spec.

kaelig: Cool with me. The v1 sticks to the spec.

We should probably keep that technique in o-grid for now (and violate to the spec), because it's being required by a lot of different modules and I think we'd see an explosion in file size (38KB each time).


For the record here's the bit in o-grid that prevents it from being output multiple times:

```scss
@if $o-grid-is-silent == false {
	@include oGridSurfaceCurrentLayout;
	@include oGridGenerate;

	// Turn silent mode back on to avoid outputting the grid twice
	$o-grid-is-silent: true;
}
```

AlbertoElias: I'm with @kaelig here, and I think we're using this technique in other modules as well.

PR #116, opened 2016-06-02
Enabling Right hand side checkboxes and option buttons. Link: https:/…

Riyanat:
…/github.com/Financial-Times/o-forms/issues/106


alicebartlett:
Thanks for this, looks good!

PR #123, opened 2016-09-27
Add design guidelines file

onishiweb:
Initial addition of design guidelines file to test new Build Service feature.

This new file will eventually expand to include adding demos in specific places amongst the documentation as well as images etc. once those features get added to the Registry.

### DIFF
--- a/designguidelines.md
+++ b/designguidelines.md
@@ -2,14 +2,20 @@ These components are used as the building blocks for creating all types of forms
 
 ## Form rules:
 
-*Labels* - Always label the form component ie a text input, select box to help the user understand it’s meaning.
-*Order the components logically* - Reflect the natural flow of a conversation. For example would you ask for someones bank details before you asked for their name?
-*Group related information* - For example Personal Details, and Payment Details.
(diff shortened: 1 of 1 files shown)


designguidelines.md
```diff
@@ -0,0 +1,15 @@
+These components are used as the building blocks for creating all types of forms. These forms should be able to be used on all devices ie desktop, mobile and tablet.
+
+## Form rules:
+
+*Labels* - Always label the form component ie a text input, select box to help the user understand it’s meaning.
```
JakeChampion: Is this meant to be a list of items? Github shows it as a paragraph of text. https://github.com/Financial-Times/o-forms/blob/7bf92adcb97af6f11c088b5cbadada3ab64337b7/designguidelines.md

JakeChampion: I think each item needs to be prepended with `- ` in order for Markdown to recognise it as a list.

onishiweb: It's not meant to be a list. Just a series of lines. Maybe I should just add a newline between each

PR #127, opened 2016-11-04
Rename vars to match others

JakeChampion:


JakeChampion:
I was aiming for internal consistency between the variable name and where it was being used, I was struggling to remember snake case for variable names and bem for class names. Happy to revert the variable names back.

onishiweb:
I think it's ok to keep this, it seems to make understanding variables easier in this instance so that's a good thing. But I don't think it should be convention as most components aren't this complicated.

PR #135, opened 2017-01-19
Add form validation 

onishiweb:
Adds the ability to validate forms based on the Constraint Validation API in browsers, applying the appropriate `o-forms` classes when an error has occurred and preventing Safari from submitting a form if errors are present (other browsers do this as standard).

Also updates readme to fit Origami structure.

### DIFF
--- a/src/js/forms.js
+++ b/src/js/forms.js
@@ -13,7 +13,8 @@ class Forms {
 
 		// o-forms should only be registered against a <form>
 		// element. If not, return to prevent errors
-		if (! this.FormEl instanceof HTMLFormElement) {
+		if (!(this.FormEl instanceof HTMLFormElement)) {
+			this.destroy();
 			return;
 		}
 
@@ -26,7 +27,7 @@ class Forms {
 			// Safari reports the validity state, but doesn't
 			// prevent form submits, so this listens to submits and
 			// checks the inputs are valid before submission.
-			this.FormEl.addEventListener('submit', this.checkAllInputs.bind(this), false);
+			this.FormEl.addEventListener('submit', this.validateForm.bind(this), false);
 
 			// All other browsers will report each item invalid on
 			// submit and prevent a form submission.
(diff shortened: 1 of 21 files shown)


onishiweb:
Still todo:

- [ ] Write 1 more test to make coverage 100%
- [ ] Fix linting errors
- [ ] Document

onishiweb:
ok, so this is ready to review! 

We haven't been able to correctly see what's going on with the 1 missing line/branch/statement that is not being hit by the tests, so I'm going to leave it for now and maybe spend a bit more time on it on Monday, but for now... go go go!

src/js/forms.js
```diff
@@ -0,0 +1,127 @@
...
+		this.FormEl = FormEl;
+		this.validFormEls = [
+			HTMLFormElement,
+			HTMLInputElement,
+			HTMLSelectElement,
+			HTMLButtonElement,
+			HTMLTextAreaElement
+		];
+
+		const declaredTestEvent = this.FormEl.getAttribute('data-o-forms-test-event');
```
JakeChampion: What is a test event? Is this used to aid in testing?

src/js/forms.js
```diff
@@ -0,0 +1,127 @@
...
+		if (!rootEl) {
+			rootEl = document.body;
+		}
+		if (!(rootEl instanceof HTMLElement)) {
+			rootEl = document.querySelector(rootEl);
+		}
+		if (rootEl instanceof HTMLElement && /\bo-forms\b/.test(rootEl.getAttribute('data-o-component'))) {
+			return new Forms(rootEl, opts);
+		}
+		return [].map.call(rootEl.querySelectorAll('[data-o-component="o-forms"]'), rootEl => new Forms(rootEl, opts));
```
JakeChampion: `Array.from` can be use here instead of `[].map.call`, I reckon it would aid in readability of the code.
`Array.from` accepts two parameters, first is an iterable (E.G. A NodeList) to turn into an array, and the second parameter is a mapping function to run on every item in the new array.

`Array.from(rootEl.querySelectorAll('[data-o-component="o-forms"]'), rootEl => new Forms(rootEl, opts))`

src/js/forms.js
onishiweb: Test event is the name I've given to the event on which to run the validation checks - explained in the readme, either `blur` or `submit`. It's a terrible name, happy to take suggestions...

src/js/forms.js
onishiweb: Ok sure, I'll add that to here and to the `component-boilerplate` literally the only bit of code that's been copied from anywhere ;)

PR #147, opened 2017-06-30
hide errortext until error modifier is used on the form

JakeChampion:
potentially breaking change but looks like most origami users are already doing this inside their own css


notlee:
@JakeChampion This is good but I'd feel better if it went out in a major release, inadvertently hiding error messages would lead to some frustrating experiences. I trust your judgements though if you two agree on a patch. :)

PR #149, opened 2017-07-03
Use arrow-down instead of arrows-up-down - Fixes #145

### LINKED ISSUE #145: change select box icon
Change the icon in all select boxes to use `arrow-down` instead of `arrows-up-down`

JakeChampion:
Old:
screen shot 2017-07-03 at 17 16 11

New:
screen shot 2017-07-03 at 17 15 47


alicebartlett:
I think the small form element might need a bit more padding..
screen shot 2017-07-04 at 10 08 36
The space should probably visually match the space to the left of the label - not sure if technically it does match, but it looks to me like it doesn't so it should look like it does, regardless of the actual pixels

Issue #143, opened 2017-06-22
No icons for dropdown menus in IE11


JakeChampion:
![bs_win10_ie_11 0](https://user-images.githubusercontent.com/1569131/27430831-000248a6-5742-11e7-9f17-0c3426fc2375.jpg)

alicebartlett:
@ajwl this would be a good one for you to work on for your bootcamp!

PR #153, opened 2017-09-25
Fix bug in ie11 that hid icons in input fields

ajwl:
The down arrow indicating a drop-down menu inside an input field was invisible on IE11, as raised by this issue: https://github.com/Financial-Times/o-forms/issues/143

This fix makes it visible on IE and other browsers so it appears as it should: 
screen shot 2017-09-25 at 16 22 18


ajwl:
👌 thanks

PR #154, opened 2017-09-26
Update codebase for obt7

JakeChampion:


src/js/forms.js
```diff
@@ -99,7 +97,7 @@ class Forms {
 		this.validateInput(input);
 	}
 
-	handleClickEvent(event) {
+	static handleClickEvent(event) {
```
notlee: >Expected 'this' to be used by class method 'handleClickEvent'.

@JakeChampion I followed your lead on making these static methods to get around the linting issue. It seems sensible but as it effects the api of an oForm instance it is technically a breaking change. I'm not worried in this case but thought if you hadn't realised it was worth mentioning incase it is an issue.

JakeChampion: Very good point, we should instead disable the eslint error for these methods and create an issue to move these to static in the next major

test/forms.test.js
```diff
@@ -158,6 +158,25 @@ describe("Forms", () => {
...
+			input.value = 'test';
+
+			formEl.addEventListener('submit', (event) => {
+				event.preventDefault();
+				throw new Error('Expected form to not be submitted');
+			}, false);
+
+			setTimeout(() => {
+				done();
+			}, 1000);
```
notlee: This could probably be reduced from 1000 to speed up the Circle build?

notlee:
Looks like pa11y tests are failing. Did they previously or do we have stricter obt7 settings?
`html > body > div:nth-child(36) > div > div > button`

JakeChampion:
Not stricter settings, they failed for a while now on o-forms

Issue #160, opened 2017-10-06
 o-forms v2 update


draysonandstock:
**Have updated the main component sketch file (here: https://drive.google.com/open?id=0B8f0KVx7ey79NlNwREpXSjJKV28)** 

_A few updates to the overall o-forms and new updated designs for Prefix-Suffix field types (more aligned with the other changes we made as part of NEXT)._ 

- [x] This includes increasing field height from 38px to 40px. This then ensures the associated buttons can be the default secondary larger buttons and align perfectly. 

- [ ] To add space between the associated check boxes and buttons to aid with touch targets on mobile. 

- [ ] Increase the smaller text to 16px across o-forms. 

- [ ] Increase the vertical space between individual fields to 20px and decrease the space between field title and entry box to 8px. This is to make it easier to see which title is attached to which entry box. 

- [x] Closing up the gap between radios and checkboxes and the associated text from 16px to 12px. This helps with then they are displayed as a vertical list on smaller breakpoints and makes it easier to know what text it is linked to. 

- [ ] To reduce the (subsequent) paragraph space on Checkboxes With Long Text to neaten up how this looks when there are a number of permissions (something that may come up more with GDPR). 

- [x] To add an optional label to any field titles so we don’t have to mark all required fields with asterisks.

notlee:
Updates have been released in v5 https://github.com/Financial-Times/o-forms/pull/180

PR #170, opened 2018-01-15
James design changes (most of them)

notlee:
Most of [James' design changes](https://github.com/Financial-Times/o-forms/issues/160). Please confirm changes against the Sketch design attached to the issue as the design changes written on the issue are not super clear. 

This is intended to be a **minor** however HTML changes will be needed to achieve the refined spacing around checkboxes/radios. Without the additional markup current spacing should be retained.

Some mixins, including some which have been around a while, are not documented in the readme. I think this is probably a sound, active decision so not to overcomplicate the documentation -- e.g. `oFormsLabelLong`. Thoughts?

## Todo

- Prefix/suffix changes still to come.

## Overview

- Reduces space between checkboxes/radios and their legend to 8px.
- Adds a wrapper class for checkboxes/radios `.o-forms__checkboxes`/`.o-forms__radios` which stacks them by default and ensures correct vertical spacing. The modifier `.o-forms__checkboxes--inline`/`.o-forms__radios--inline` can be used to inline the inputs (this is the exception).  **Markup change required to receive benefits.** 
- Increases field height from 38px to 40px, so associated buttons can be the default secondary larger button height.
- Adds class to indicated an optional field `.o-forms__label--optional`.
- Adds class `.o-forms__label--long` for lengthy labels (i.e. t&cs).
- Tweak font sizes.
- Introduces `o-normalise` to the demos.
- Merges "Checkboxes With Long Text" demo within the main checkbox demo.

## Checkboxes before/after

screen shot 2018-01-15 at 16 30 19

screen shot 2018-01-15 at 16 33 22

##  Text Input before/after

Relatively unchanged, outside of `o-normalise` which is now in the demos. New "optional input" demo.
screen shot 2018-01-15 at 17 29 59

## Optional Input Closeup
screen shot 2018-01-15 at 17 01 06


src/scss/mixins/_labels.scss
```diff
@@ -1,10 +1,9 @@
 /// Label styles
 @mixin oFormsLabel {
-	@include oTypographySansBold(1);
+	@include oTypographyBold('sans');
```
gvonkoss: Just curious, why the change?

src/scss/mixins/_labels.scss
```diff
@@ -16,18 +15,31 @@
 
 /// Additional info
 @mixin oFormsAdditionalInfo {
-	@include oTypographySans(0);
+	@include oColorsFor(o-forms-field-additional-info body, text);
 	display: block;
-	color: oColorsGetColorFor(o-forms-field-additional-info body, text);
+	font-size: $_o-forms-font-size-small;
```
gvonkoss: Same, why the change?

src/scss/_variables.scss
```diff
@@ -15,16 +20,29 @@ $o-forms-is-silent: true !default;
...
+$_o-forms-typography-scale-small: oTypographyGetScale(-1);
+$_o-forms-typography-scale-regular: oTypographyGetScale(0);
+$_o-forms-typography-scale-large: oTypographyGetScale(1);
+
+$_o-forms-font-size-default: oFormsGetFontSize($_o-forms-typography-scale-regular); // 16px
+$_o-forms-font-size-small: oFormsGetFontSize($_o-forms-typography-scale-small); // 14px
+$_o-forms-font-size-large: oFormsGetFontSize($_o-forms-typography-scale-large); // 18px
+
+$_o-forms-spacing: oTypographySpacingSize(3); // 12px
+$_o-forms-checkbox-ledgend-spacing: oTypographySpacingSize(2); // 8px
```
gvonkoss: typo?

README.md
```diff
@@ -60,25 +60,55 @@ Each form field is made up of at least 3 elements:
...
-	<label for="a">Unchecked (default)</label>
-	<input id="a" type="radio" class="o-forms__radio" checked="checked" />
-	<label for="a">Checked</label>
-</div>
+<fieldset class="o-forms">
+	<label class="o-forms__label">Radio controls</label>
+	<div class="o-forms__radios">
+		<input id="a" type="radio" class="o-forms__radio" />
+		<label for="a">Unchecked (default)</label>
+		<input id="a" type="radio" class="o-forms__radio" checked="checked" />
```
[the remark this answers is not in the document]
notlee: Good spot!

src/scss/message.scss
```diff
@@ -6,14 +6,14 @@
 /// Messages styles
 @mixin oFormsMessage($class: 'o-forms') {
 	.#{$class}__message {
+		@include oColorsFor(o-forms__wrapper-highlight, background);
 		max-width: $_o-forms-field-max-width;
 		padding: 0 $_o-forms-padding;
-		margin-bottom: 16px;
+		margin-bottom: $_o-forms-spacing;
```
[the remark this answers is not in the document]
notlee: A I didn't realise that, good shout. I'll leave this as is however as those mixins delegate to `oTypographySpacingSize` anyhow.

src/scss/mixins/_labels.scss
notlee: @gvonkoss 
Alice did this as:
>The rest of the font props are being set elsewhere so this only needs
to set the correct weight which can be done with this mixin.
https://github.com/Financial-Times/o-forms/pull/170/commits/7326e0a79f3b91b716088b27f3f9efdc71540685

I see `oFormsCommonFieldBase` sets the font-family. I think it should be using `oTypographySans` but don't think we can switch to that without it being breaking as `$o-forms-font-family` is a public variable. If you agree that's how it should be, we could create an issue set the font family with `oTypographySans`?

src/scss/mixins/_labels.scss
notlee: We wanted to change the font size here. This has been updated to use `oTypographySize` based on our previous conversation.

src/scss/mixins/_common.scss
```diff
@@ -5,19 +5,18 @@
 
 /// Common field styles
 @mixin oFormsCommonFieldBase {
-	@include oTypographyMargin($top: 3);
+	@include oColorsFor(o-forms-field-standard);
+	margin-top: $_o-forms-spacing;
```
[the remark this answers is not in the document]
notlee: 🤔 as discussed...

Seemingly arbitrary padding like this can be swapped out for a size intune with our vertical rhythm. In this case the height of inputs are set to 40px anyway:
https://github.com/Financial-Times/o-forms/pull/170/files#diff-73b5b8a1f62da0fc9ef102b351d0fee9R34

notlee:
Sorry @gvonkoss, I fixed a bug and made a couple design changes since you approved after discussion with James. Mind double checking the commits since?

As discussed:
- Error spacing between checkboxes.
- Size of t&c (removed `o-forms__label--long`)
- Bugfix for overflowing stacked checkboxes https://github.com/Financial-Times/o-forms/pull/170/commits/ac1ae0e92c21e28291e832606d428c9728e860f0.
screen shot 2018-01-17 at 17 20 10

gvonkoss:
Good bug spot :+1: It's still approved :shipit:

PR #171, opened 2018-01-19
Remove prefix -  update suffix - checkbox control

notlee:
- Fixes spacing and examples of toggle checkboxes.
- Removes prefix entirely.
- Simplifies and updates suffix uses.
- Updates icon/spacing for select dropdowns and checkboxes.
- Fixes `o-forms__checkbox--align-right` renaming to `o-forms__checkbox--right`. Adds to demo.

**Suffix / Prefix before & after**
screen shot 2018-01-19 at 17 19 15
screen shot 2018-01-19 at 17 18 46

**icon/spacing before & after**
screen shot 2018-01-19 at 17 21 12
screen shot 2018-01-19 at 17 22 12

**align right**
`--right` will be used instead of the suffix to display a show/hide password form control (discussed with @draysonandstock )
screen shot 2018-01-19 at 16 45 08


README.md
```diff
@@ -257,7 +258,7 @@ The `oForms` mixin also allows customisation of the base classname:
 
 - `oFormsWrapper` - an area around the form, which can be coloured.
 - `oFormsMessage` - styles for global form messages.
-- `oFormsPrefixSuffix` - adds ability to affix information/buttons to form fields.
+- `oFormsSuffix` - adds ability to affix information to form fields.
```
gvonkoss: Is there an example of what 'information' is?

src/scss/mixins/_common.scss
```diff
@@ -98,15 +95,38 @@
...
 		#{$checkbox} + #{$label} {
 			width: 100%; // We do not use display:block to maintain consistent margins (not collapsed).
 			box-sizing: border-box;
+			margin: $_o-forms-checkbox-legend-spacing 15px $_o-forms-spacing 0; // margins do not collapse, making the space between checkbox/radio inputs a little larger
+			&:last-of-type {
+				margin-right: 0;
+			}
+		}
+
+		$align-right: '--right';
```
gvonkoss: doesn't this technically align left ...?

notlee: @gvonkoss 🤔 it aligns the checkbox input to the right of the label, as the modifier is applied to the input I think this makes sense?

gvonkoss: oh I _see_.
I thought we were referring to the whole of that section being aligned to the bottom of the input 🤦‍♀️ 
As in the `show password  []` is the suffix that we'd be aligning, not the actual checkbox in relation to the text. 

Would that also be an example of 'information' you can affix as a suffix? 😬

notlee: This is no longer in a suffix, it's just a checkbox (aligned right to the label) under an input
35163073-66cb0a4a-fd3d-11e7-960c-6884cc7d239c
Does that make sense? We might want to change the markup/demos to be clearer.

README.md
notlee: 🤔 The only design condoned suffix currently is a button so I'll swap out 'information' to 'button'... technically though you could put what you like in that space.

PR #175, opened 2018-01-25
Begin refactoring.

notlee:
WIP

- Renames `o-forms__checkbox-toggle` to `o-forms__toggle`
- Refactors to move class names outside of mixins where possible (e.g. radio checkbox)
- Consistency outputs CSS in `base.scss` rather than a file by file basis.
- Removes ability to label class names.
- Removes deprecated code.


src/scss/base.scss
```diff
@@ -27,37 +29,102 @@
...
 
-	@include oFormsRadioCheckbox();
+	/// Radio and checkboxes
+
+	.#{$class}__checkboxes,
+	.#{$class}__radios {
+		@include oFormsRadioCheckboxContainer();
+	}
+
+	.#{$class}__radio,
```
gvonkoss: I would argue that there is potential for this to raise issues, the difference between `radio` and `radios` may be too subtle, perhaps we may need to change the element/modifier situation, or the name of the wrapper?

gvonkoss: and also could we not refer to the wrapper of radios and checkboxes as something that wraps 'options'? Or 'choices' or 'selection' or something like that?

src/scss/base.scss
```diff
@@ -27,37 +29,102 @@
...
 
+	/// Message
+
+	.#{$class}__message {
+		@include oFormsMessage();
+	}
+
+	/// Wrapper
+
+	.#{$class}__wrapper {
```
gvonkoss: 🤔 I don't know what this is wrapping....? (just by reading this code)

src/scss/message.scss
```diff
@@ -4,41 +4,31 @@
...
-		padding: 0 $_o-forms-padding;
-		margin-bottom: $_o-forms-spacing;
-		margin-left: $_o-forms-padding * -1;
-		margin-right: $_o-forms-padding * -1;
-		color: oColorGetColorFor('o-forms__message-default', text);
+@mixin oFormsMessage() {
+	@include oColorsFor(o-forms__wrapper-highlight, background);
+	@include oTypographySize($_o-forms-typography-scale-large);
+	max-width: $_o-forms-field-max-width;
+	padding: 0 $_o-forms-padding;
```
gvonkoss: these are not `oTypographyMargin/Padding` worthy?

notlee: This is horizontal spacing which uses the grid gutter :)

notlee:
@gvonkoss good points regarding naming, as discussed this will come in a alter PR 👍

PR #181, opened 2018-01-31
Input text vertical spacing

notlee:
James spotted an alignment issue. The text within a form input is not quite centred. Oddly (I don't know why) this can be avoided by not setting a height. Instead we will rely on padding to produce the correct height.

Wait for it...

kapture 2018-01-31 at 12 01 37


notlee:
Looks like this can also be solved by setting `line-height` to the height of the field, but I'll leave as is I think.

PR #183, opened 2018-02-02
Radios styled as button set

jonnywyatt:
A button set allowing either one selected by default, or none.

Design from @draysonandstock

image

### DIFF
--- a/demos/src/radios-button-styling.mustache
+++ b/demos/src/radios-button-styling.mustache
@@ -1,0 +1,45 @@
+<fieldset class="o-forms">
+    <legend class="o-forms__label">No default selection</legend>
+    <div class="o-forms__radio-buttons-group">
+        <input type="radio" name="radio4" value="daily" class="o-forms__radio-buttons__button" id="radio41" />
+        <label for="radio41" class="o-forms__label">Daily</label>
+        <input type="radio" name="radio4" value="weekly" class="o-forms__radio-buttons__button" id="radio42" />
+        <label for="radio42" class="o-forms__label">Weekly</label>
+    </div>
+</fieldset>
+
+<fieldset class="o-forms">
+    <legend class="o-forms__label">Default selected</legend>
+    <div class="o-forms__radio-buttons-group">
+        <input type="radio" name="radio5" value="daily" class="o-forms__radio-buttons__button" id="radio51"
+               checked="checked" />
+        <label for="radio51" class="o-forms__label">Daily</label>
+        <input type="radio" name="radio5" value="weekly" class="o-forms__radio-buttons__button" id="radio52" />
+        <label for="radio52" class="o-forms__label">Weekly</label>
+        <input type="radio" name="radio5" value="monthly" class="o-forms__radio-buttons__button" id="radio53" />
+        <label for="radio53" class="o-forms__label">Monthly</label>
+    </div>
+</fieldset>
+
+<fieldset class="o-forms o-forms--error">
+    <legend class="o-forms__label">Error</legend>
+    <div class="o-forms__radio-buttons-group">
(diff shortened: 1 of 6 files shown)


src/scss/mixins/_radio-checkbox.scss
```diff
@@ -133,12 +133,95 @@
...
+  display: inline-block;
+  padding-left: $_o-forms-radiocheckbox-default-size + $_o-forms-label-padding-left; // radio/checkboxes are being styled as pseudo-elements, which means we have to account for the width of the radio/checkboxes when adding padding to the labels
+  position: relative;
+  cursor: pointer;
+  font-weight: oFontsWeight('normal');
+  padding-top: 2px; // To align the text with the center of the checkbox/radio
+  user-select: none;
+}
+
+@mixin oFormsRadioButtonsStyled($selectors: (legend: '.o-forms__radio-buttons__legend', wrapper: '.o-forms__radio-buttons__wrapper', label: '.o-forms__label', radio: '.o-forms__radio-buttons__button')) {
```
notlee: In v5 we removed the ability to rename these selectors with mixin parameters as it was implemented inconsistently.
https://github.com/Financial-Times/o-forms/blob/ca76910f71c5aadf599c0ed4c05d441fc5c6582f/src/scss/mixins/_radio-checkbox.scss#L44

src/scss/mixins/_radio-checkbox.scss
```diff
@@ -133,12 +133,95 @@
...
+
+  #{$legend},
+  #{$radio} {
+    @include oNormaliseVisuallyHidden;
+  }
+
+  $sidePadding: 32px;
+
+  #{$label} {
+    @include oFormsLabel;
```
notlee: `@include oFormsLabel;` will be inherited with the cascade, so I think this can be deleted.

demos/src/radios-button-styling.mustache
```diff
@@ -0,0 +1,57 @@
+<div class="o-forms">
+    <p>No default selection</p>
+    <fieldset class="o-forms__radio-buttons">
```
notlee: https://github.com/Financial-Times/o-forms/blob/master/demos/src/radios.mustache#L1

Maybe something along these lines?
```
<fieldset class="o-forms">
    <legend class="o-forms__label">No default selection</legend>
    <div class="o-forms__radio-buttons-group">
        <input type="radio" name="radio4" value="daily" class="o-forms__radio-buttons__button" id="radio41" />
        <label for="radio41" class="o-forms__label">Daily</label>
        <input type="radio" name="radio4" value="weekly" class="o-forms__radio-buttons__button" id="radio42" />
        <label for="radio42" class="o-forms__label">Weekly</label>
    </div>
</fieldset>
```
`o-forms` is used as a containing element for the field. And the legend becomes the field label visually speaking.

`o-forms__radio-buttons__wrapper` would then become `o-forms__radio-buttons-wrapper`. What would you think about calling it `o-forms__radio-buttons-group` for consistency with the new classes just added in v5?
https://github.com/Financial-Times/o-forms/blob/master/demos/src/radios.mustache#L4

With these changes the `o-forms__radio-buttons` class and styles aren't needed (`o-forms` removes the default legend styling)
https://github.com/Financial-Times/o-forms/pull/183/files#diff-12b0210cae4dd2ece978ad72b68c87c2R152

notlee: Without any specific wrapper, this feature could then be output along these lines:
https://github.com/Financial-Times/o-forms/blob/master/src/scss/base.scss#L91

jonnywyatt: Great suggestions thanks. All done.

notlee:
@jonnywyatt approves 🎉

notlee:
@jonnywyatt approves 🎉

PR #184, opened 2018-02-02
Update radios button styled.

notlee:
Builds on https://github.com/Financial-Times/o-forms/pull/183/files

- Adds the ability to optionally output styled radio buttons with `oFormsRadioButtonsStyledFeature` rather than include them in base features.
- Uses `o-forms__group` for margins consistent with standard checkboxes/radios.
- Adds `o-forms__group--inline-together`, uses floats and grids for visually improved browser support.
- Renames `o-forms__radio-buttons__button` `o-forms__radio-button`
- Adds to readme.


notlee:
Hey @jonnywyatt, thanks for making those changes in the previous PR. I've built on that with a few extra tweaks. What do you think?

Think we're good to go after this! :) Other than we have used inconsistent indentation which `obt verify` doesn't seem to have flagged. I'll have a look at that Monday!

notlee:
@jonnywyatt approves 🎉

PR #188, opened 2018-02-14
Styled radio buttons - allow +/- choice; spaced styling

jonnywyatt:
Design from @draysonandstock 

image

### DIFF
--- a/demos/src/radios-button-styling.mustache
+++ b/demos/src/radios-button-styling.mustache
@@ -9,6 +9,16 @@
 </fieldset>
 
 <fieldset class="o-forms">
+	<legend class="o-forms__label">Positive / negative choice with a default selection</legend>
+	<div class="o-forms__group o-forms__group--inline-together">
+		<input type="radio" name="radio8" value="yes" class="o-forms__radio-button" id="radio81" />
+		<label for="radio81" class="o-forms__label">Yes</label>
+		<input type="radio" name="radio8" value="no" class="o-forms__radio-button o-forms__radio-button--negative" id="radio82" checked />
+		<label for="radio82" class="o-forms__label">No</label>
+	</div>
+</fieldset>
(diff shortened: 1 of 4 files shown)


notlee:
I've had a little think about your questions regarding `.o-forms__group` specificity issues in relation to our public mixins setup. I think the idea was that group wrappers remain relatively generic and enforce spacing around grouped labels (maybe the legend should be providing bottom margin rather than the group providing top margin but that's another issue. 😬 ). A side effect of having taken this approach is that it doesn't make it easy to support styling on the group that's specific to its contents without introducing more modifiers -- which also means implementers of the component get visually breaking changes unless they add the new modifier. 

With the changes so far we've lost the spacing enforced by the group element. It might be better to try to avoid modifying it for specific contents if possible.

screen shot 2018-02-14 at 13 53 23
screen shot 2018-02-14 at 13 53 43

Maybe a less disruptive approach to avoid such issues would be to use `:before` pseudo elements with a margin to apply the new border look?

jonnywyatt:
This component doesn't need the vertical spacing set by the group element. Using a pseudo-element to set the border doesn't feel like the simplest thing to do. It feels like the cause of the issue is that the generic class o-forms is prescribing some styling (max-width, margin) that isn't applicable to this case. So I should probably add a variant of o-forms which is more generic? Although if we were starting from scratch you could argue that o-forms should be generic and we add a more specific variant

notlee:
As discussed, it's not ideal the group wrapper assumes each child is a label which works as an individual entity. We could  improve this in the next major by having specific wrappers which do not concern themselves with spacing of surrounding elements. For now to avoid a breaking markup change we can do something more along these lines:
https://github.com/Financial-Times/o-forms/compare/styled-radio-buttons-updates-psydo?expand=1

PR #205, opened 2018-04-13
Inline forms and other visual improvements.

notlee:
## Updates
1. Adds inline forms (IE9 falls back to stacked).
2. Updates styled radio button error state.
3. Reduces styled radio button padding. 
4. Removes white background from radio buttons.

## Notes to self (and the curious)
1. There are some mad selectors and specificity here -- feedback welcome. Inputs/input groups have unique margins between their label/legend, i.e. a group of checkboxes or radios are closer to their legend than styled radio buttons or text inputs are. This means the inputs/input groups own the margin (margin-top) rather than the labels/legends (margin-bottom). Input groups such as checkboxes may also have negative margins which enable their children to be inline within a stacked form. As the margins are unique and sometimes negative to support inline child elements it becomes problematic to remove the input/input group margin with a global inline form modifier. **Identifying the type of form input at the top most level e.g. `.o-forms.o-forms--textarea` might help increase flexibility and decrease specificity**, unfortunately requiring this would be a breaking.
2. IE10-11 Support grid but do not support @supports, so a browser hack has been used to only reset input margins if grid is supported or the browser is IE10/11.
3. Legends prevent `display: grid` (at least in many browsers),  [they establish a new block formatting context](https://html.spec.whatwg.org/multipage/rendering.html#the-fieldset-and-legend-elements). This is why the inline container was added to wrap the legend contents with a css grid `legend.o-forms.o-forms--inline > .o-forms__inline-container`. If a legend is not being used however their is no need to use the inline container `div.o-forms.o-forms--inline`.

## Screenshots

**Styled radio buttons with reduced padding and  a new error state.**
localhost_8999_demos_local_radios-button-styled html

**Radios without a white background.**
localhost_8999_demos_local_radios html

**Inline form in IE10**
bs_win8_ie_10 0

**Inline form in Chrome**
localhost_8999_demos_local_inline html 2

### DIFF
--- a/demos/src/inline.mustache
+++ b/demos/src/inline.mustache
@@ -149,7 +149,7 @@
     <div class="o-forms__errortext">Please enter a valid url</div>
 </div>
 
-<div class="o-forms o-forms--inline o-forms--textarea">
+<div class="o-forms o-forms--inline">
     <div class="o-forms__inline-item">
 	    <label for="o-forms__textarea" class="o-forms__label">Textarea</label>
 	    <div class="o-forms__additional-info">Prompt text follows the label if required.</div>
@@ -157,7 +157,7 @@
 	<textarea id="o-forms__textarea" class="o-forms__textarea"></textarea>
 </div>
 
-<div class="o-forms o-forms--inline o-forms--textarea o-forms--error">
+<div class="o-forms o-forms--inline o-forms--error">
 	<label for="o-forms__textarea-invalid" class="o-forms__label">Textarea (error)</label>
 	<textarea id="o-forms__textarea-invalid" class="o-forms__textarea"></textarea>
     <div class="o-forms__errortext">Invalid entry</div>
--- a/demos/src/inline.mustache
+++ b/demos/src/inline.mustache
@@ -1,118 +1,94 @@
 <fieldset class="o-forms o-forms--inline">
     <div class="o-forms__inline-container">
-        <div class="o-forms__inline-item">
-            <legend class="o-forms__label">No default selection</legend>
+        <legend class="o-forms__label">No default selection</legend>
+        <div class="o-forms__group o-forms__group--inline-together">
+            <input type="radio" name="radio4" value="daily" class="o-forms__radio-button" id="radio41" />
+            <label for="radio41" class="o-forms__label">Daily</label>
+            <input type="radio" name="radio4" value="weekly" class="o-forms__radio-button" id="radio42" />
+            <label for="radio42" class="o-forms__label">Weekly</label>
         </div>
-        <div class="o-forms__inline-item">
-            <div class="o-forms__group o-forms__group--inline-together">
-                <input type="radio" name="radio4" value="daily" class="o-forms__radio-button" id="radio41" />
-                <label for="radio41" class="o-forms__label">Daily</label>
-                <input type="radio" name="radio4" value="weekly" class="o-forms__radio-button" id="radio42" />
-                <label for="radio42" class="o-forms__label">Weekly</label>
+</fieldset>
+
+<fieldset class="o-forms o-forms--inline">
+    <div class="o-forms__inline-container">
+        <legend class="o-forms__label">Positive / negative choice with a default selection</legend>
+        <div class="o-forms__group o-forms__group--inline-together">
+            <input type="radio" name="radio8" value="yes" class="o-forms__radio-button" id="radio81" />
+            <label for="radio81" class="o-forms__label">Yes</label>
+            <input type="radio" name="radio8" value="no" class="o-forms__radio-button o-forms__radio-button--negative" id="radio82" checked />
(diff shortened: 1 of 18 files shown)


jonnywyatt:
Epic PR, I'll set aside some time!
Thanks for this. Go Origami!

notlee:
Styled radio button error text appeared to have a slightly larger margin compared to radios/checkboxes because the later overflowed their container, making them closer to the error. Just tweaked that.

src/scss/mixins/_radio-checkbox.scss
```diff
@@ -119,28 +121,47 @@
...
+                margin-bottom: 0;
+            }
         }
     }
 
+    // Inline radios/checkboxes.
+    // Half inline margins as inline margins do not collapse.
+    $group-item-verticle-spacing: $_o-forms-group-spacing / 2;
     &--inline {
+        margin: #{-$group-item-verticle-spacing + $_o-forms-checkbox-legend-spacing} #{-$_o-forms-checkbox-horizontal-spacing} #{-$group-item-verticle-spacing} 0;
```
jonnywyatt: minor, but typo in 'vertical'

demos/src/inline.mustache
```diff
@@ -0,0 +1,165 @@
...
+    <div class="o-forms__inline-item">
+	    <label for="o-forms-standard" class="o-forms__label">Text input</label>
+	    <div class="o-forms__additional-info">Prompt text follows the label if required..</div>
+    </div>
+	<input type="text" id="o-forms-standard" class="o-forms__text" required />
+</div>
+
+<div class="o-forms o-forms--inline o-forms--error">
+	<label for="o-forms-invalid" class="o-forms__label">URL input :invalid</label>
+    <input type="url" id="o-forms-invalid" class="o-forms__text" value="hp://invalid.com" />
```
jonnywyatt: I found that aria-describedby on the input, referencing the error message, will cause it to be read when the input is focussed.

notlee: Ahh this is very nice. I'll update all the demos.

jonnywyatt: It works the same for you? 
Try it in My account - enable controlCentrePreferencesPage flag then hit ft.com/preferences, edit Personal details, blank the First name field and hit enter

notlee: Have you got the same approach to work with fieldset/legend?

src/scss/mixins/_radio-checkbox.scss
```diff
@@ -119,28 +121,47 @@
...
 
-    input + .#{$class}__label {
-        width: 100%; // We do not use display:block to maintain consistent margins (not collapsed).
-        box-sizing: border-box;
-        margin: $_o-forms-checkbox-legend-spacing $_o-forms-checkbox-horizontal-spacing $_o-forms-spacing 0; // margins do not collapse, making the space between checkbox/radio inputs a little larger
-        &:last-of-type {
-            margin-right: 0;
+    // Stacked radios/checkboxes.
+    // @breaking Remove the not selectors by removing the need to override label margins.
+    &:not(#{&}--inline-together):not(#{&}--inline) {
```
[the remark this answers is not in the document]
notlee: I struggled with this as I couldn't undo the margin reset with `--inline-together` or `--inline` reliably -- so I instead only apply the margin reset if `--inline-together` or `--inline` are not present. You're right this might cause problems later though, would be great to avoid! I'm open to suggestions though if you can think of a better way without a breaking change.

demos/src/inline.mustache
notlee: I'm not totally sure I understand your question. It works great for me using VoiceOver on My Account, and the technique works for us in our demos where the error is displayed.

jonnywyatt: 👍

src/scss/mixins/_radio-checkbox.scss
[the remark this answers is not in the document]
jonnywyatt: Is it possible to not set default margins, instead only set them on the variants where they're needed? Then you avoid the need to reset them in specific cases.
I'm aware this might be hard to do in practise without a bunch of breaking changes. Just a thought.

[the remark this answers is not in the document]
notlee: I'm not sure this is possible without a breaking as we share a class names across groups of checkboxes/radios with a group of styled radios. 
`.o-forms__group {} // stacked checkboxes or radios`
`.o-forms__group .o-forms__group--inline {}  // inline checkboxes or radios`
`.o-forms__group .o-forms__group--inline-together // styled radio buttons`

demos/src/inline.mustache
notlee: I ended up adding aria-describedby to the demo for additional content but not errors. But I've noted it on a separate issue for errors. 👍

PR #209, opened 2018-04-23
Update vertical spacing.

notlee:
- Checkbox/radio groups now have a 12px margin from their legend (up from 8px).
- Checkboxes/radios within a group now have a 16px margin between one another (down drom 20px).

I've also renamed some variables to avoid confusion. Unfortunately we have a class `o-forms__group` which does not relate to `$_o-forms-group-spacing`.

`$_o-forms-group-spacing` is the space between instances of `o-forms` or [form sections](https://www.ft.com/__origami/service/build/v2/demos/o-forms@5.5.1/sections). I've renamed `$_o-forms-group-spacing` to `$_o-forms-section-spacing` and `$_o-forms-inline-group-spacing` to `$_o-forms-inline-section-spacing`.

before
screen shot 2018-04-23 at 18 00 35

after
screen shot 2018-04-23 at 18 05 34

### DIFF
--- a/src/scss/_variables.scss
+++ b/src/scss/_variables.scss
@@ -30,11 +30,12 @@ $_o-forms-field-default-font-size: 14px;
 /// Spacing variables.
 /// @access private
 /// @type String
-$_o-forms-group-spacing: oTypographySpacingSize(5); // 20px
-$_o-forms-inline-group-spacing: oTypographySpacingSize(8); // 32px
 $_o-forms-spacing: oTypographySpacingSize(3); // 12px
-$_o-forms-checkbox-legend-spacing: oTypographySpacingSize(2); // 8px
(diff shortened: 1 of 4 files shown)


src/scss/_variables.scss
```diff
@@ -30,11 +30,12 @@ $_o-forms-field-default-font-size: 14px;
 /// Spacing variables.
 /// @access private
 /// @type String
-$_o-forms-group-spacing: oTypographySpacingSize(5); // 20px
-$_o-forms-inline-group-spacing: oTypographySpacingSize(8); // 32px
```
JakeChampion: I think renaming variables makes this a breaking change

notlee: They're marked private (sassdoc and underscore) so a rename is okay (in theory).

PR #212, opened 2018-04-24
Controls

notlee:
Review ignoring whitepace: https://github.com/Financial-Times/o-forms/pull/212/files?w=1

- Adds inline controls.
- Adds a `bleed` option to remove form padding.

E.g.
screen shot 2018-04-24 at 17 58 08

To achieve:
screen shot 2018-04-24 at 17 58 49

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,17 @@ If `o-forms` is used on a `fieldset` a wrapper `o-forms__inline-container` must 
 </fieldset>
 ```
 
+Using `o-forms--inline-controls` instead of `o-forms--inline` keeps the inputs inline even on mobile. It also expands the label forcing the inputs to align right. This is useful for long lists of simple styled radio buttons which are used as controls, e.g. repeated yes/no options. For an example see [the registry](https://www.ft.com/__origami/service/build/v2/demos/o-forms/inline-controls).
+
+```html
+<fieldset class="o-forms o-forms--inline-controls">
+    <div class="o-forms__inline-container">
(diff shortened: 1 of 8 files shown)


notlee:
@jonnywyatt we might want to update error margins? Currently they're the same as for checkboxes and radios.

jonnywyatt:
Request from @draysonandstock  - could you remove the top margin from the error box so it matches the gap for input and textarea errors?

jonnywyatt:
Also, o-forms--inline-controls seems to carry a bottom margin of 32px, can it be set to zero, or the default?

jonnywyatt:
One more, the custom class needs passing through to the mixin here - https://github.com/Financial-Times/o-forms/blob/master/src/scss/base.scss#L145

README.md
```diff
@@ -63,6 +63,17 @@ If `o-forms` is used on a `fieldset` a wrapper `o-forms__inline-container` must
 </fieldset>
 ```
 
+Using `o-forms--inline-controls` instead of `o-forms--inline` keeps the inputs inline even on mobile. It also expands the label forcing the inputs to align right. This is useful for long lists of simple styled radio buttons which are used as controls, e.g. repeated yes/no options. For an example see [the registry](https://www.ft.com/__origami/service/build/v2/demos/o-forms/inline-controls).
```
gvonkoss: As discussed, the 'even' in this sentence is a little confusing, could we say something like, 'keeps the inputs inline from mobile viewports upwards` ?

PR #217, opened 2018-05-11
Add form status

notlee:
Adds the concept of a form status including "saving" and "saved".

status

screen shot 2018-05-11 at 15 39 03

Also works with inline controls:
screen shot 2018-05-11 at 15 39 20

### DIFF
--- a/src/scss/base.scss
+++ b/src/scss/base.scss
@@ -198,6 +198,11 @@
 /// @param {string} $class - The block level class name (BEM naming convention).
 /// @output Styles for form status such as "saving" and "saved".
 @mixin oFormsStatusFeature($class: 'o-forms') {
+	// Icons align to a 40x40 pixel grid with 10px padding on each side.
+	// In this case however we want to remove that padding to align the icon flush with the input.
+	// https://github.com/Financial-Times/fticons/blob/master/contributing.md#design
+	$status-icon-size: $_o-forms-field-small-height;
+	$status-icon-padding: ($status-icon-size/40) * 10;
 	@include oBrandConfigureFor($component: 'o-forms') {
 		// Base status styles.
 		.#{$class}__status {
@@ -207,13 +212,13 @@
 			white-space: nowrap;
 			align-items: center;
 			min-width: 5em;
-			margin-top: 1px; // Aligns status with error text for seemless replacement.
 			@include oTypographySize($_o-forms-typography-scale-small);
 			&:after {
 				content: '';
 			}
 			&:before {
 				content: '';
+				margin: #{-$status-icon-padding} 0 #{-$status-icon-padding} #{-$status-icon-padding};
 			}
 		}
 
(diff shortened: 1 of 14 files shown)


notlee:
@jonnywyatt I found with Voiceover when the `aria-hidden` attribute is set to `false` the status is read aloud because of `aria-live` (it would be great if we could test other screenreaders)

notlee:
Spacing on inline controls with an error also updated. Screenshot: 

screen shot 2018-05-11 at 16 48 30

src/scss/base.scss
```diff
@@ -189,6 +194,78 @@
...
+				width: 10px;
+				height: 10px;
+				border-width: 2px;
+			}
+		}
+
+		// Saved status.
+		.#{$class}--saved .#{$class}__status {
+			&:after {
+				content: 'Saved';
```
jonnywyatt: I'm confused why it's setting content with a pseudo-element, can't the label sit inside the status element?

jonnywyatt:
lgtm

src/scss/base.scss
notlee: It could, but this way helps keep language consistent across apps and makes it a little easier to implement as you don't have to worry about updating the copy between states.

src/scss/base.scss
```diff
@@ -189,6 +194,78 @@
...
 }
+/// @access public
+/// @param {string} $class - The block level class name (BEM naming convention).
+/// @output Styles for form status such as "saving" and "saved".
+@mixin oFormsStatusFeature($class: 'o-forms') {
+	// Icons align to a 40x40 pixel grid with 10px padding on each side.
+	// In this case however we want to remove that padding to align the icon flush with the input.
+	// https://github.com/Financial-Times/fticons/blob/master/contributing.md#design
+	$status-icon-size: $_o-forms-field-small-height;
+	$status-icon-padding: ($status-icon-size/40) * 10;
```
gvonkoss: What is the benefit over making this calculating v setting the variable to 10px?

src/scss/base.scss
```diff
@@ -189,6 +194,78 @@
...
+		.#{$class}__status[aria-hidden="false"] {
+			display: flex;
+			visibility: visible;
+			opacity: 1;
+		}
+
+		// Saving status.
+		.#{$class}--saving .#{$class}__status {
+			&:after {
+				content: 'Saving';
```
gvonkoss: Just to be clear — the request for this feature is explicitly for a 'Saving' scenario? Would it need to handle different wording in the future?

notlee: Different status may be introduced but I think that can be handled in this way too. We don't currently have a need for dynamic or translated content.

src/scss/base.scss
gvonkoss: 👍

src/scss/base.scss
gvonkoss: 👍

PR #218, opened 2018-05-14
Form Status JS

notlee:
- Add a method to set form input status.
- Visually hide icon only statuses so screenreaders can read them.
- Set inactive statuses to `display: none` when not inline so they take no space

### DIFF
--- a/src/js/forms.js
+++ b/src/js/forms.js
@@ -98,6 +98,7 @@ class Forms {
 		const oFormsEl = input.closest('.o-forms');
 		oFormsEl.classList.add('o-forms--error');
 		oFormsEl.classList.remove('o-forms--valid');
+		Forms.updateInputStatus(input, null); // Remove any form status e.g. "saving" or "saved".
 	}
 
 	handleInvalidEvent(event) {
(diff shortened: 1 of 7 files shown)


notlee:
I will follow this up with tests in the morning.

src/js/forms.js
```diff
@@ -134,6 +135,59 @@ class Forms {
 		this.FormEl = undefined;
 	}
 
+	static updateInputStatus (input, status) {
+		if (!(input instanceof HTMLElement)) {
+			throw new TypeError(`Expecting "input" to be an instance of "HTMLElement" but instead found ${oFormsEl instanceof HTMLElement}.`);
```
rowanmanning: I don't _fully_ understand this error. So will it output the following?

```
Expecting "input" to be an instance of "HTMLElement" but instead found true.
```

notlee: Hah that is a mistake. I'll update `${oFormsEl instanceof HTMLElement}` to `${typeof oFormsEl}`

JakeChampion: For any input which is not an HTMLElement instance, this would error with:
>Expecting "input" to be an instance of "HTMLElement" but instead found false.

The "but instead found false" could make a user think that they passed in `false`, which they may not have done. 
Perhaps something like:
>`Expected input (${input}) to be of type "HTMLElement".

src/js/forms.js
```diff
@@ -134,6 +135,59 @@ class Forms {
 		this.FormEl = undefined;
 	}
 
+	static updateInputStatus (input, status) {
+		if (!(input instanceof HTMLElement)) {
+			throw new TypeError(`Expecting "input" to be an instance of "HTMLElement" but instead found "${typeof oFormsEl}".`);
```
JakeChampion: 👍

Issue #219, opened 2018-05-29
Saved feedback - pushes following content downwards


jonnywyatt:
When the saving / saved label appears, could it occupy the existing space and not alter the height of the whole Yes / No block? Then it won't push any following content downwards.

draysonandstock:
![wheat box position 1](https://user-images.githubusercontent.com/11214093/40652904-8a48c4a8-6331-11e8-8c2b-d6397a707ca9.png)
![wheat box position 2 has shifted by the height of the feedback message](https://user-images.githubusercontent.com/11214093/40652905-8a5e2848-6331-11e8-9d29-335ae329ca94.png)

notlee:
Addressed in [v5.7.2](https://github.com/Financial-Times/o-typography/releases/tag/v5.7.2)

PR #220, opened 2018-06-01
Make status sit within o-forms bottom margin.

notlee:
So that setting a status does not push content down.
See issue: https://github.com/Financial-Times/o-forms/issues/219

kapture 2018-06-01 at 12 17 13


notlee:
@jonnywyatt @Financial-Times/origami-core This negative margin feels like a gross choice but may be the best approach. Alternatives welcome :)

PR #231, opened 2018-10-23
Add date capture input group.

notlee:
- Add date capture input group.
- Also remove some private typography variables.

Discussed this design in the uxd surgery with Mark H and Simon C. 

screen shot 2018-10-23 at 16 10 41

### DIFF
--- a/src/scss/_variables.scss
+++ b/src/scss/_variables.scss
@@ -7,25 +7,14 @@
 $o-forms-is-silent: true !default;
 
 /// Field font family variable.
+/// Defaults to the sans font defined for o-typography.
 ///
 /// This is required to override default system form element styling,
 /// since `<input>`, `<textarea>`, `<select>`… don't inherit global font styles
 ///
 /// @access public
 /// @type String
-$o-forms-font-family: MetricWeb, sans-serif !default;
-
-/// Typography scale variables.
-/// @access private
-/// @type Number
-$_o-forms-typography-scale-small: -1;
-$_o-forms-typography-scale-regular: 0;
-$_o-forms-typography-scale-large: 1;
-
-/// Font size variables.
-/// @access private
-/// @type String
-$_o-forms-field-default-font-size: 14px;
+$o-forms-font-family: $o-typography-sans !default;
 
 /// Spacing variables.
 /// @access private
@@ -39,7 +28,7 @@ $_o-forms-checkbox-legend-spacing: oTypographySpacingSize(3); // 12px
 $_o-forms-checkbox-horizontal-spacing: 15px;
 $_o-forms-checkbox-vertical-spacing: oTypographySpacingSize(4); // 16px
 $_o-forms-padding: oGridGutter();
-$_o-forms-default-error-text-margin: oTypographySpacingSize(1); // 4px;;
+$_o-forms-default-error-text-margin: oTypographySpacingSize(1); // 4px;
 
 /// Field dimention variables.
 /// @access private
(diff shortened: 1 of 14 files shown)


src/scss/_variables.scss
```diff
@@ -39,7 +28,7 @@ $_o-forms-checkbox-legend-spacing: oTypographySpacingSize(3); // 12px
 $_o-forms-checkbox-horizontal-spacing: 15px;
 $_o-forms-checkbox-vertical-spacing: oTypographySpacingSize(4); // 16px
```
gvonkoss: possibly unrelated, but as a pedantic comment, it looks like, of the values that are multiples of 4, half of the variables use `oTypographySpacingSize` and the others don't ... ?

src/scss/mixins/_date.scss
```diff
@@ -0,0 +1,33 @@
+@mixin oFormsDateFeatures($class) {
+    .#{$class}--date .#{$class}__group {
+        // We want to use `box-sizing: border-box` to set a max-height which includes padding.
+        // However when setting width to an `ex` value we want padding to be in addition.
+        $padding-width: $_o-forms-field-default-padding-leftright * 2;
+        margin-top: oTypographySpacingSize(2);
+
+        .#{$class}__label {
+            @include oFormsGroupLabel();
+            margin-bottom: oTypographySpacingSize(1);
```
gvonkoss: Also a pedantic observation: some properties are assigned variable values, others are using this o-type function

src/scss/mixins/_date.scss
```diff
@@ -0,0 +1,33 @@
...
+        .#{$class}__text,
+        .#{$class}__label {
+            width: calc(2.5ex + #{$padding-width});
+            margin-right: $_o-forms-spacing;
+            float: left;
+        }
+
+        .#{$class}__text--year,
+        .#{$class}__label--year {
+            width: calc(5ex + #{$padding-width});
```
gvonkoss: Do we use `ex` anywhere else? It seems a bit unusual to introduce it to this component, and to the dates in this component alone 🤔

src/scss/_variables.scss
notlee: Erm some aren't based on typography size directly (like the ones which should align to button height, and others are horizontal rather than vertical spacing (but arguably should use typography spacing).

src/scss/mixins/_date.scss
notlee: Hopefully the distinction is variables for shared values and o-type directly for non-shared values. Whether this is true or not I'm unsure. 🙈Maybe something for a different PR.

gvonkoss: If the distinction is meant to be implied, I think we should probably document that somewhere because we're not too consistent about it 😬

src/scss/mixins/_date.scss
notlee: As they're inline the width for these inputs is a little different. The width indicates what kind of data we're expecting, 2 chars for dd/mm or 4 chars for yyyy. But come to think of it should I be using `ch` rather than `ex` here. 🤔 GDS do the latter.

notlee: Okay I've used to use `ch` (despite the new commit message saying the opposite). https://github.com/Financial-Times/o-forms/pull/231/commits/d04422e558741b634037fd0e3d9d06bbb33c0ea6#diff-53809cd931a4317a557310479fbe30ceR23

src/scss/mixins/_date.scss
notlee: Yeah good shout. There're an awful lot of spacing variables already. Rationalising is something for another PR though... It would be great to do a breaking of o-forms and pull out some things into their own components to simplify things, in my opinion.

PR #235, opened 2018-11-22
Correct right aligned checkbox checked state.

notlee:
Before
screenshot 2018-11-22 at 14 40 21

After
screenshot 2018-11-22 at 14 38 06

This visual bug was introduced here: https://github.com/Financial-Times/o-forms/pull/230


JakeChampion:
We should revisit the conversation about visual regression testing

Issue #236, opened 2018-12-04
V7


gvonkoss:
## Justifications for the breaking change

- o-forms through  the build service is very big
- documentation can be challenging because there's a lot in here
- There are a number of different elements that have different needs—such as theming for toggle buttons or styled radio buttons
- there are very many mixins and they can get complicated
- o-forms class on an individual field has confused people
- specificity is an issue, we use the + selector with labels and inputs
- markup hasn't been changed but features have been added, which has lead to bloat, e.g. the layout styles for GDPR
- remove class name customisation ([accepted proposal](https://github.com/Financial-Times/origami-proposals/issues/4))
- potentially remove 'sections'

## Discussed solutions so far ("splitting" means new repo, new component):

- Breaking up `o-forms` entirely: Split out each type or category of input into a component, and create a 'form layout' component to put other components into the alignment in the newly created grid(s)

- Split out each type of input into a component, but maintain `o-forms` as a 'macro'-component which which pulls in all of the separate components and serves them as current `o-forms` does

- Less splitting, maybe just split out the more complex parts of forms, e.g. toggle buttons.

- No splitting, keep everything in `o-forms` (after evaluating usefulness of each input type)

<details>
<summary>Existing form elements:</summary>

- Text Input
  - with prompt
  - as 'optional'
  - acceptance state
  - error state
  - disabled state
  - full container width
  - narrow area (small)
- Select Box
  - with prompt
  - acceptance state
  - error state
  - disabled state
  - multi select
  - narrow area
- Text area
  - with prompt
  - as 'optional'
  - error state
  - disabled state
  - full container width
- Date input (three text inputs)
  - with prompt
- *Round* Radio buttons
  - with prompt
  - inline
  - error state
  - disabled state
- *Rectangular* Radio Buttons
  - no prompt?
  - no selection
  - default selection
  - different positive/negative styling
  - error state
  - disabled state
- Anchors styled as radio buttons (naming? what is this for?)
  - default selection only
- Checkboxes
  - with prompt
  - inline
  - right aligned under text input
  - error state
  - disabled state
  - with long text
  - with wide text
- 'Toggle' checkbox
  - default
  - inverse
- Suffix (button? anchor? icon?)
  - regular text box
  - regular select box
  - small text box
  - small select box
- Sections (was this up for debate?)
  - with label and form element
  - as message only
  - full width
  - error state
- Layout for all
  - text and element stacked (default)
  - text and element inline (elements stacked)
[...]

PR #239, opened 2019-01-08
V7: Style overhaul

### LINKED ISSUE #194: IE10 Select Box Icon
screen shot 2018-03-14 at 11 00 15

### LINKED ISSUE #202: Remove `$o-forms-font-family`.
Remove `$o-forms-font-family` in favour of `o-typography` mixins instead, e.g. `oTypographySans`.
https://github.com/Financial-Times/o-forms/blob/master/src/scss/_variables.scss#L16

### LINKED ISSUE #224: Consider adding o-buttons dependency
`o-forms` matches `o-buttons` variables. On the next breaking consider adding an official dependancy on `o-buttons` to share these variables and keep them in sync. Alternatively, consider a new dependent for `o-buttons` and `o-forms`.

E.g. Input height matches button height. Hover transition is shared (o-visual-effects?). Styled radio buttons match button padding and icon size.
https://github.com/Financial-Times/o-buttons/blob/v5.13.2/scss/_variables.scss#L35
https://github.com/Financial-Times/o-buttons/blob/v5.13.2/scss/_icon.scss#L73
https://github.com/Financial-Times/o-buttons/blob/61271d11f575eaab6c5d3b924ea9304c5596fffa/scss/_mixins.scss#L40

### LINKED ISSUE #237: Remove default padding left/right from .o-forms
Users have to reset this depending on the context they're placing the form.

gvonkoss:
# V7 Styles
#### Concept ⁿ
This PR has a number of new changes which are partially a reflection of the [initial rewrite](#original-concept).

After a lot of back and forth, I've finally managed to get a structure to support flexbox that I'm content with. As an overview, all elements fit into the following 'grid' :

_(if we ever want to edit this ascii—I highly recommend doing in an editor because it looks real weird in plaintext)_
```
                                            Notes on alignment
                                             - titles and inputs stack by default
                                               e.g.
                                               <title>
                                               <input>
Flex 'Grid':                                   <input>
                                               
┌— parent flex container ———————————————┐  ——→ parent container is responsible
|      (one of div or label)            |      for inlining the title and the inputs
|                                       |      whether they are a group or individual
|  ┌— flex child for titles —————————┐  |      e.g.
|  |    (group titles or             |  |      <title>  <input>
|  |        individual input labels) |  |               <input>
|  └—————————————————————————————————┘  |               
|                                       |
|  ┌— flex child for inputs —————————┐  |  ——→ input flex child can align its
|  |    (group inputs or             |  |      inputs independently of the parent
|  |        individual inputs)       |  |      e.g.
|  └—————————————————————————————————┘  |      <title>
└———————————————————————————————————————┘      <input> <input>
                                               
                                             - the above can be combined to fully
                                               inline the title and input
                                               e.g.
                                               <title>  <input> <input>

```
Based on this, the classes have shifted as well;
- The parent element accepts twill be responsible for setting the base of the form elements (e.g. typography, inverse, required).
- The title flex child is responsible for distributing styling based on what kind of text exists within that title (e.g. main title, prompt, required text)
- The input flex child is responsible for setting  the modifiers on its input elements (e.g. radio, textarea, &c). This also involves modifiers like small (for text/select), status (for toggle), and validity (in/valid).

Both the flex container and input flex child receive individual inline modifiers so that the behaviour described above is possible. 

This layout allows a large amount of flexibility in terms of what and how much of it goes where, meaning the form elements will hold their basic shape if a modifier is/n't present.

I have addressed the previous comments about the original concept as best as I could, and have removed the older code I had doubts about. This means there is now more specific references to elements within these containers through classes, and 'double' modifiers have been split out.
Custom classes are not allowed _within_ element styles, but will wrap elements (only radio and anchors). There is only *one* mixin for customisation now, and it can handle any icon that we have in the image set.

The suffix will stay but it doesn't require the dependency of o-buttons, which will likely close #224.
Sections have been removed, and we will put `o-message` forward as a suitable substitute.

~`optional` has been swapped out for `required`—the justification being that it seems to be more common in terms of from UX that a field is marked as required _and_ it reflects the HTML accurately (`<input required>`). This is also going to be discussed with design. ~ We've decided to stick to 'optional'

*Demos*
I've restructured the demo templates so that we don't need to repeat the markup and change every instance of it—however, since it is a rather complex component and every type of input has a few variations, I've had to include a JSON file for each type of input. 

Hopefully the structure of these JSON files make sense, they should be the same for multiple inputs or a single input. 

There are two base templates for now (`multiple-input-field`,   `single-input-field`), but this should pave the way, partially, for us to implement `o-forms` demos in React when `oat` is ready ✌️ 

### TODO
- [x] Shared variables/values
- [x] Reach an agreement on class names
- [x] Fix 'checkbox  right'
- [x] Fix inverse 'required' text colour
- [x] Test with screen reader
- [x] aria-fy
- [x] `<input type='password'>`
- [ ] `<input type='file'>`
- [x] Pa11y
- [ ] Migration
- [ ] README
- [ ] JavaScript
- [ ] Tests

### Issues that this PR covers so far.
Closes #194 
Closes #202 
Closes #237 

#### Original concept
<details>
<summary>Details about the initial changes to V7 (out of date)</summary>
This is the start of the V7 major. It focuses mainly on updating the markup and the styling. However, there are aspects of `o-forms` that are unclear, meaning that there are a few parts of the component that haven't been covered in this PR, which still require discussion. 

#### Markup

The markup has been simplified. This doesn't mean that it is shorter, or that there is less of it, but that less information is required for each form element (e.g. less `class`, `id`, `for`, etc attributes).

These changes **do not** cover aria attributes, yet. They **do** implement a 'trickle down' class name system, where the modifying classes are mainly (but not exclusively) set on the parent form element (e.g. `<label>`, `<fieldset>`). This also means that the HTML _must_ be accurate, or the styles will not apply.

All of the HTML has been validated. There is another point up for discussion, however—sections without headings (makes for a validation 'warning', not error).

<details>
<summary> Text Input Example </summary>

```diff
-<div class="o-forms">
-	<label aria-describedby="text-box-info" for="o-forms-standard" class="o-forms__label">Text input</label>
-	<div id="text-box-info" class="o-forms__additional-info">Prompt text follows the label if required..</div>
-	<input type="text" id="o-forms-standard" class="o-forms__text" required />
-</div>
+<label class="o-forms o-forms--text-input">
+	<span class="o-forms__label">Text Input</span>
+	<span class="o-forms__prompt">Optional prompt text</span>
+	<input type="text" name="text" value="">
+</label>
```
</details>

<details>
<summary> Checkboxes Example </summary>

```diff
-<div class="o-forms">
-	<legend aria-describedby="default-checkboxes-info" class="o-forms__label">Default checkboxes</legend>
-	<div id="default-checkboxes-info" class="o-forms__additional-info">Prompt text follows the ledgend if required.</div>
-	<div class="o-forms__group">
-		<input type="checkbox" name="checkbox" value="1" class="o-forms__checkbox" id="checkbox11" checked="checked" />
-		<label for="checkbox11" class="o-forms__label">Checkbox 1</label>
-		<input type="checkbox" name="checkbox" value="2" class="o-forms__checkbox" id="checkbox12" />
-		<label for="checkbox12" class="o-forms__label">Checkbox 2</label>
-		<input type="checkbox" name="checkbox" value="3" class="o-forms__checkbox" id="checkbox13" />
-		<label for="checkbox13" class="o-forms__label">Checkbox 3</label>
-	</div>
-</div>
+<fieldset class="o-forms o-forms--checkbox">
+	<legend>
+		<span class="o-forms__label">Default checkboxes</span>
+		<span class="o-forms__prompt">Optional prompt text</span>
+	</legend>
+	<label>
+		<input type="checkbox" name="checkbox" value="1" checked/>
+		<span>Stacked 1</span>
+	</label>
+	<label>
+		<input type="checkbox" name="checkbox" value="2"/>
+		<span>Stacked 2</span>
+	</label>
+	<label>
+		<input type="checkbox" name="checkbox" value="3" checked/>
+		<span>Stacked 3</span>
+	</label>
+</fieldset>
```
</details>

#### Styles

Currently, the styling works across all browsers, down to IE9. There may be some exceptions to this rule, as I didn't check every version of every browser on every operating system. 

This is possible because most CSS grid and flexbox applications have been removed, which is the justification for the `<span>` tags within `<legend>`s and `<label>`s.

The `wide` variant has been removed entirely, as has the `max-width` for a form element. Form elements now span the width of their container.

The classes have been changed and reduced:

<details>
<summary> Parent-Level Classnames </summary>

(the removed class names are a combination of parent and child classes):
```diff
o-forms
+o-forms--anchor
+o-forms--checkbox
+o-forms--checkbox--extra-content
+o-forms--checkbox--right
o-forms--date
-o-forms--error
-o-forms__group
-o-forms__group--inline-together
o-forms--inline
+o-forms--inline-children
-o-forms__inline-container
-o-forms--inline-controls
+o-forms--invalid
-o-forms--loading
+o-forms--optional
+o-forms--radio--box
+o-forms--radio--round
+o-forms--select
+o-forms--saved
o-forms--saving
+o-forms--small
+o-forms--text-input
+o-forms--text-area
-o-forms__toggle
+o-forms--toggle
-o-forms__toggle--inverse
+o-forms--toggle--inverse
+o-forms--valid
-o-forms--wide
```
</details>

<details>
<summary> Child Classnames </summary>

```diff
-o-forms__additional-info
-o-forms__checkbox
-o-forms__checkbox--right
+o-forms__current
+o-forms__error-message
-o-forms__errortext
o-forms__label
-o-forms__label--day
-o-forms__label--month
-o-forms__label--year
-o-forms__label--optional
+o-forms__prompt
-o-forms__text
-o-forms__text--day
-o-forms__text--month
-o-forms__text--small
-o-forms__text--valid
-o-forms__text--year
-o-forms__textarea
-o-forms__toggle
-o-forms__toggle__checkbox
-o-forms__toggle--inverse
-o-forms__radio
-o-forms__radio-button
-o-forms__radio-button--current
-o-forms__radio-button--negative
+o-forms--radio--button--negative (need to fix this one)
-o-forms__select
-o-forms__select--small
o-forms__status
o-forms__status--icon-only
-o-forms__status--left
+o-forms__status--stacked
```
</details>

#### Mixins

Almost all mixins have been removed. There are three public facing mixins now:
```
oForms();
oFormsCustom();
oFormsCustomIcon();
```
Each have a set of arguments, the map for `oForms()` being particularly long 🤔.

### Discussion points
There is a file `src/scss/TBD` (and its counterpart `demos/src/TBD`) that has a few sass mixins that I need clarification on @notlee.

It also has the `section` and `suffix` styles, which we've briefly mentioned may be deprecated, but I'm not sure who to ask, or if there is any more information on that, so that is another aspect that needs to be revisited. 

### Missing / caveats

**Every variable under the sun:** I've haphazardly added the `o-forms-is-silent` variable so that I could use it, but there are a lot of fixed numbers across styles that I will add variables for next.

**Definitive class names:** I've settled on box-styled and round-styled for now. I'm aware of the descriptive choice not being ideal, but I disagree with calling anchors 'styled radio buttons'. If the argument is that it is unsafe to use descriptive classnames, I disagree in this case because it is a direct reflection of the styles. Other options that have been put forward are 'switch' or 'semantic controls'. I'm unsure. 

**Separation of concerns:** I've coupled inputs that look similar for the sake of sharing mixins. E.g. text inputs, textarea, select, dates, all fall under 'input', checkboxes and radios fall under 'controls'. These are private.

**Functionality:** I've not touched the JavaScript of this component yet, ~so the tests will likely be failing~ linting is important 🤦‍♀️  

**Dependencies**: ~`o-loading` [v3](https://github.com/Financial-Times/o-loading/pull/19) needs to be released~ in order to update the dependency, same with `o-typography` [v5.8.0](https://github.com/Financial-Times/o-typography/pull/166). Also, the addition of `o-buttons` may be irrelevant if we decide to remove suffixes / make `o-spacing`.

**Pa11y, Migration, README:** I've not updated any of these, and won't until we've merged this PR, or are further down the line.

### Justification
Despite the many options we presented in #236 most of it is hampered by browser support (we can't use a grid system to slot form elements into) and general style sharing. It is particularly difficult to separate 'categories' of inputs, so to speak, so it may be confusing to split `o-forms` into smaller categorical components in a semantic way. 

I've settled for rewriting the whole of the component and using the flexibility of the primary mixin to control the bundle size. Hopefully. 
</details>


main.scss
```diff
@@ -1,40 +1,110 @@
...
+		@include _oFormsRadioBox($disabled, $invalid);
+	}
+
+	@if index($elements, 'select') {
+		@include _oFormsSelect(
+			$disabled,
+			$small,
+			$inline-label,
+			$invalid,
+			$valid);
```
rowanmanning:
```suggestion
			$valid
		);
```

src/scss/_checkbox.scss
```diff
@@ -0,0 +1,124 @@
...
+@mixin _oFormsCheckbox(
+	$additional-content: null,
+	$disabled: null,
+	$inline-children: null,
+	$inline-label: null,
+	$invalid: null,
+	$right: null
+) {
+	.o-forms--checkbox {
+		label span {
```
rowanmanning: The `label` I'm fine with because there can only be one per input in valid HTML, but the `span` selector could be problematic if someone needs to add in any amount of custom stuff over the top of o-forms, or even if we later decide we need a new feature and a span happens to be the right element for that. Wonder whether we should adopt a class like `o-forms__label-text`?

(this is in more places, just commented on the first one I found)

src/scss/_checkbox.scss
```diff
@@ -0,0 +1,124 @@
...
+		}
+
+		@if $invalid {
+			&.o-forms--invalid .o-forms__error-message {
+				@include oTypographyMargin($top: 2);
+			}
+		}
+
+		@if $right {
+			&.o-forms--checkbox--right {
```
rowanmanning: Should this be `o-forms--checkbox-right`? It's a single BEM modifier that's _kind_ of masquerading as two. Can a modifier have a modifier of its own? Just the first instance I found but saw a few more in the demos, worth a discussion maybe

src/scss/_checkbox.scss
gvonkoss: I would argue that element specificity is low enough for a new class to be applied to a span and take precedence? 🤔

src/scss/_anchor-box.scss
```diff
@@ -0,0 +1,37 @@
+/// @access private
+/// @output Styling for anchor 'controls'
+
+@mixin _oFormsAnchorBox {
+	.o-forms--anchor {
```
notlee: I don't think it makes sense to call this just `anchor`, but a good name is hard. ` o-forms--anchor-radio-box`, ` o-forms--pseudo-radio-box`, ` o-forms--fake-radio-box`?

src/scss/_checkbox.scss
notlee: I concur with Rowan's feedback 👍 Relying on users to override styles we apply (and may change in minor or patch releases) is unreliable.

src/scss/_custom.scss
```diff
@@ -0,0 +1,132 @@
...
+			display: block;
+		}
+	}
+}
+
+/// @access public
+/// @param {String} $icon-name Name of the icon to request from the image set
+/// @param {String} $icon-class Class name for custom icon
+/// @param {Map|null} $theme Custom theme map
+@mixin oFormsCustomIcon($icon-name, $icon-class, $theme: null) {
```
notlee: Isn't this specif to the custom anchored radio button? Maybe `oFormsCustom` could accept `$opts` which could include a list of icons for the anchored radio button?

src/scss/_custom.scss
```diff
@@ -0,0 +1,132 @@
...
+	$theme: null
+) {
+
+	@if not $input {
+		@error('Please specifiy whether you would like to customise anchor or radio inputs');
+	}
+
+	$theme: _oFormsThemeToBrandVariants($theme);
+
+	.#{$custom-class} {
```
notlee: In `o-buttons` we talked about creating custom modifier classes but not allowing totally custom classes through a mixin interface. How would you feel about `o-forms--mycustom.o-forms--anchor`? I'm still unsure on this though. Maybe  something to discuss in person with the team?

src/scss/shared/_controls.scss
```diff
@@ -0,0 +1,74 @@
...
+			}
+		}
+	}
+}
+
+/// @access private
+/// @output Shared pseudo element styles for checkboxes, toggle, round- and box-style radio buttons.
+@mixin _oFormsControlsPseudoElements {
+	content: '';
+	cursor: pointer;
```
notlee: Looks like a pointer used to be applied to the labels of checkbox and radio buttons too.

src/scss/_toggle.scss
```diff
@@ -0,0 +1,107 @@
...
+		@include _oFormsToggleInverse();
+	}
+
+
+}
+
+/// @access private
+/// @output styles for inverse toggle based on a checkbox input
+@mixin _oFormsToggleInverse {
+	.o-forms--toggle--inverse {
```
notlee: Much better inverse style 👏 Might I suggest `.o-forms--toggle.o-forms--inverse`? We could then add support for the `.o-forms--inverse` class on other inputs as and when needed.

src/scss/_status.scss
```diff
@@ -0,0 +1,77 @@
...
+		}
+	}
+
+	.o-forms--saving .o-forms__status {
+		&:after {
+			content: 'Saving';
+		}
+
+		&:before {
+			@include oLoadingContent($opts: ('color': 'dark', 'size': 'mini'));
```
notlee: The loading icon isn't showing because `color` is `theme` now.

notlee: Oh wait I got this 😁

```suggestion
			@include oLoadingContent($opts: ('theme': 'dark', 'size': 'mini'));
```

src/scss/_date.scss
```diff
@@ -0,0 +1,57 @@
...
+		clear: both;
+
+		label {
+			@include oTypographyMargin($bottom: 0);
+			float: left;
+		}
+
+		@if $invalid {
+			&.o-forms-input--invalid {
+				@include _oFormsInputValidity('invalid', $element);
```
rowanmanning: Would it be possible to output some of these base styles at the top level, inside the `oForms` mixin? Right now we're duplicating a load of output CSS for each input type.

If we have a top-level `.o-forms-input--invalid` selector which this then expands on, then we'd reduce quite a lot I think. The same goes for `.o-forms-input--valid`

gvonkoss: Turns out we rely in on that class to apply validity to an input over the `:invalid` / `:valid` pseudo-selectors.

There's no straightforward way (as far as I have seen) of adding those pseudo selectors to the markup, and the browser engine will apply them automatically. E.g. if an input is `:required`, all required inputs would load as invalid until they are filled out.

We also need the individual styles to be specific about which element to  affect under what condition—otherwise the styles could apply to anything under the `--invalid` or `--valid` parent class, as they are broader, such as `color` or `background-color`

gvonkoss: Having said that, I could modify `input`, `select` and `textarea` and remove the `$invalid`, `$valid` options permanently 🤔

PR #253, opened 2019-04-11
V7: Validation

gvonkoss:
This PR handles form validation.

I've asked a few products how they validate their forms, and there is a mix of native browser validation and product specific overrides.

Bearing this in mind, this PR handles a few situations:
- Individual input fields will be validated on the `blur` event (e.g. when focus is removed from an input)
- Individual input fields will be re-evaluated on the `input` event (e.g. when an invalid field is being corrected)
- If native browser validation is active:
  - Any form `novalidate` attributes will be removed
  - All input fields within the form will be validated on the form's native `submit` event.
- If native browser validation is **not** active:
  - `novalidate` will be added to the form element(s)
  - All input fields within the form will be validated on an `input[type=submit]` click

The native browser validation has a number of benefits:
- It will automatically scroll the invalid field into view
- It generates a message that is specific to the reason behind the invalid state
- The messages that are generated are specific to the language set by the user for the browser

As far as I can tell, the reason _not_ to use native validation is primarily because the message tool tips do not get any type of styling—and we have no intention of overriding the styles for these tool tips

Whether or not native validation is used, the browser will still apply the pseudo-element states of `:valid` or `:invalid` to the inputs, which I've leveraged for our styling.

I've gone further with the styling for invalid/valid states—they are now automatically incorporated to the base styles of a form. They've been written to affect any input that can accept `:invalid`/`:valid` — as it stands, it only affects `input[type=text]`, `select`, `textarea` elements, because of the way we've chosen to style our checkboxes/radio buttons. These do not have a design, but I am waiting for a response from UXD.

This means that an input will only ever receive styling to indicate its validity if it is in fact invalid according to the [`ValidityState` API](https://developer.mozilla.org/en-US/docs/Web/API/ValidityState).

In addition to that, the form validation will only handle the `:valid` state if the input has previously been `:invalid`.

You can see the behaviour I've described above by visiting the validation html and clicking/hitting enter on the submit button and then going back to fill out some fields.

This PR does **not** handle state changes (saving/saved)—that will be separate.


src/js/forms.js
```diff
@@ -1,201 +1,99 @@
...
+		if (e.type === 'input') {
+			let field = e.target.closest('.o-forms-input');
+			if (e.target.validity.valid && field.classList.contains(this.class.invalid)) {
+				field.classList.replace(this.class.invalid, this.class.valid);
+			}
 		}
 
-		if (this.opts && this.opts.applyValidState && oFormsElExists) {
-			oFormsEl.classList.add('o-forms--valid');
+		if (e.type === 'click' || (e.type === 'keydown' && e.key === 13)) {
```
rowanmanning: Minor thing: could we use a constant for this instead of a magic number? E.g. define the `KEY_RETURN` variable, then `e.key === KEY_RETURN`

src/js/forms.js
```diff
@@ -1,201 +1,99 @@
...
 
-	validateInput(input) {
-		if (input.checkValidity() === false) {
-			this.invalidInput(input);
-			return false;
+	/**
+	 * Event Handler
+	 * @param {Object} event - The event emitted by element/window interactions
+	 */
+	handleEvent(e) {
```
rowanmanning: Not sure where to put this comment, but if you enter invalid input in one of the demo date fields then shift+tab backwards from year to month, the error state goes away

gvonkoss: I believe I've fixed it!

Issue #174, opened 2018-01-25
Improve validation error experience for screen reader users


notlee:
Suggestion by @benbarnett 

>For visually impaired users, 'rolling' (blur) validation can easily be missed when tabbing through a form. Perhaps something like:
>
>When the form is submitted...
>If the validation passes, the submission carries on as normal
>if the validation fails, an error summary is appended to the top of the form, and it is given focus to alert screen-reader users

Related Issue: https://github.com/Financial-Times/o-forms/issues/152

notlee:
`aria-describedby` may be helpful to improve this experience:
 https://github.com/Financial-Times/o-forms/pull/205#discussion_r181650719

PR #254, opened 2019-04-12
V7: A11y

gvonkoss:
This is a collection of the a11y changes/attempts I've made while shaping `o-forms` v7. 

It is slightly wordy, but there have been a lot of attempts/realisations 🙈 

It's likely that some of it will change/be added, but there seem to be a lot of caveats and I wanted to jot it down while it was still fresh in my mind.

All feedback is appreciated, I really don't want to be wrong about this!

This also relates to the suggestion in #174 ~but I don't think `o-forms` should be responsible for setting the error summary area because we have no way of determining what the error messages/forms are like for individual products—thoughts on this?~


ACCESSIBILITY.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+`aria-labelledby` and `aria-describedby` are similar attributes, though semantically they provide different functionality.
+You can [read more about them on MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute):
+
+>A label provides essential information about an object, while a description provides extended information that the user might need.
+
+#### `aria-label` & `aria-hidden`
+
+Independently of whether or not a group of inputs is in a `<fieldset>` or a `<div>`, their structure exhibits a peculiar type of behaviour: AT will focus on the `<input>`, read out its label, and then focus on the label text, and read that out again. To avoid focusing on the label to provide a more natural 'tabbing' experience, we recommend combining `aria-label` and `aria-hidden` on the grouped inputs:
```
notlee: Interesting! Is this a common practice and does it work well across screen readers?

ACCESSIBILITY.md
```diff
@@ -0,0 +1,225 @@
...
+_Note 2: `aria-hidden` has often been used to visually hide/show elements. However, it's primary functionality is to hide the contents of an element from the AT_
+
+
+#### Example
+
+Putting everything we've just seen together, our multiple input field should look like this:
+```html
+<div class="o-forms-field" role="group" aria-labelledby="example-group-title" aria-describedby="example-group-info">
+	<span class="o-forms-title">
+		<span class="o-forms-title--main" id="example-group-title">Group Title</span>
```
notlee: As above: Are these classes correct?

ACCESSIBILITY.md
```diff
@@ -0,0 +1,225 @@
...
+
+This means that invalid inputs will recieve messages tailored to describe the condition that has _not_ been met. These messages are presented in the language that the user has chosen for their browser. In addition to that, the browser will shift focus onto the first invalid input. Depending on the browser, the AT will read out the generated error message. 
+
+#### Non native validation
+
+We do _not_ style the tool tip that the browser's message appears in, so frequently forms are given the `novalidate` attribute, which doesn't run any native validation on submit. This means that the messages can be overriden and styled according to the FT's design and the beahviour determined by the product's needs.
+
+Primarily because of the above, `o-forms` doesn't make any decisions about validation in terms of behaviour.
```
notlee: Not true anymore?

ACCESSIBILITY.md
```diff
@@ -0,0 +1,225 @@
...
+
+#### `aria-live` regions
+
+At its core, `aria-live` is an attribute that can be set to an area of a document to denote that the AT should pick up changes to that area. It is possible to determine what kind of changes it should recognise (additions, changes to text, etc), how forceful it should be in informing the user of the change (polite, assertive), alongside a few more advanced features.
```
notlee: I guess this section needs updating since your work on validation and error messages? Since it handles aria-live (I'm assuming) do we need to talk about it in such detail?

ACCESSIBILITY.md
```diff
@@ -0,0 +1,225 @@
...
+`o-forms` has been tested with our [contrast checker](https://registry.origami.ft.com/components/o-colors#demo-contrast-ratio-checker) to make sure that our colour contrasts meet [WCAG](https://www.w3.org/WAI/standards-guidelines/wcag/) requirements, so this document is going to walk through `aria` attributes we recommend using.
+
+### Table of Contents:
+- [Single Inputs](#single-inputs)
+- [Multiple Inputs](#multiple-inputs)
+- [Form Validation](#form-validation)
+
+## Single inputs
+`o-forms` considers these types of input 'single input fields`  
+(Currently, they represent the input types that we provide styling for):
```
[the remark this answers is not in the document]
gvonkoss: Nah, I think we can delete this to be honest

ACCESSIBILITY.md
```diff
@@ -0,0 +1,225 @@
...
+		<textarea></textarea>
+		<!-- or -->
+		<select>...</select> 
+	</span>
+</label>
+```
+AT will read this markup correctly, by identifying the type of input declared, and the label for it. As long as our markup is semantic, this should be read out as intended. 
+
+## Multiple inputs
+`o-forms` considers these types of input 'multiple input fields` (currently, they represent the input types that we provide styling for):
```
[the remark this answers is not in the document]
gvonkoss: as above

ACCESSIBILITY.md
```diff
@@ -0,0 +1,225 @@
...
++<div class="o-forms-field" role="group">
+	...
+</div>
+```
+This denominates the `<div>` as the element that holds a group of inputs, similar to the behaviour of a `<fieldset>`.  
+[According to MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_group_role): 
+>'A group should be used to form a logical collection of items with related functionality ...'
+
+#### `aria-labelledby` & `aria-describedby`
+This attribute should also be added to the parent `<div>`. It accepts an id, which should be set on `<span class="o-forms-title--main">`:
```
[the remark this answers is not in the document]
gvonkoss: What do you mean?

ACCESSIBILITY.md
gvonkoss: nope, about to change it

ACCESSIBILITY.md
gvonkoss: it needs to be rewritten, really. On it 🏃‍♀

ACCESSIBILITY.md
```diff
@@ -0,0 +1,236 @@
...
+The base structure for all of these types of input is as follows:
+```html
+<label>
+	<span class="o-forms-title">
+		<span class="o-forms-title__main">
+			Label to describe the input here
+		</span>
+	</span>
+
+	<span class="o-forms-input ...modifiers">
```
notlee:
```suggestion
	<span class="o-forms-input">
```

How would you feel about removing this? Would like to avoid pretend markup confusing anyone.

ACCESSIBILITY.md
```diff
@@ -0,0 +1,236 @@
...
+-	<span class="o-forms-input__label">Checkbox 1</span>
++	<span class="o-forms-input__label" aria-hidden="true">Checkbox 1</span>
+</label>
+```
+
+By defining what the label of the input is and hiding the actual label from the AT, we'll make the transition between inputs far smoother. 
+
+_Note 1: `aria-label` also accepts an id if you would rather set an id on the input's label_
+
+_Note 2: `aria-hidden` has often been used to visually hide/show elements. However, it's primary functionality is to hide the contents of an element from the AT_
```
notlee: Is this saying we sometimes use `aria-hidden` as a selector to visually hide elements? Took me a while to make the connection. I don't think a user needs to think about that?

ACCESSIBILITY.md
```diff
@@ -0,0 +1,236 @@
...
+
+#### Non native validation
+
+We do _not_ style the tool tip that the browser's message appears in, so frequently forms are given the `novalidate` attribute, which doesn't run any native validation on submit. This means that the messages can be overriden and styled according to the FT's design and the behaviour determined by the product's needs.
+
+Primarily because of this, we've chosen to default to non-native validation, using [custom errors](./README.md#custom-errors) and to producing an [error summary](#error-summary) instead.
+
+#### `aria-live` regions
+
+At its core, `aria-live` is an attribute that can be set to an area of a document to denote that the AT should pick up changes to that area. It is possible to determine what kind of changes it should recognise (additions, changes to text, etc), how forceful it should be in informing the user of the change (polite, assertive), alongside a few more advanced features.
```
notlee:
```suggestion
At its core, `aria-live` is an attribute that can be set to an area of a document to denote that the AT should announce changes to that area. It is possible to determine what kind of changes it should recognise (additions, changes to text, etc), how forceful it should be in informing the user of the change (polite, assertive), alongside a few more advanced features. 
```

ACCESSIBILITY.md
```diff
@@ -0,0 +1,236 @@
...
+
+We do _not_ style the tool tip that the browser's message appears in, so frequently forms are given the `novalidate` attribute, which doesn't run any native validation on submit. This means that the messages can be overriden and styled according to the FT's design and the behaviour determined by the product's needs.
+
+Primarily because of this, we've chosen to default to non-native validation, using [custom errors](./README.md#custom-errors) and to producing an [error summary](#error-summary) instead.
+
+#### `aria-live` regions
+
+At its core, `aria-live` is an attribute that can be set to an area of a document to denote that the AT should pick up changes to that area. It is possible to determine what kind of changes it should recognise (additions, changes to text, etc), how forceful it should be in informing the user of the change (polite, assertive), alongside a few more advanced features. 
+
+If you choose not to use the error summary functionality that comes with `o-forms`, please consider using the `aria-live` attribute, keeping the following in mind:
```
notlee: Doesn't o-forms handle this automatically?

ACCESSIBILITY.md
```diff
@@ -0,0 +1,236 @@
...
+Overall, it outlines that:
+- There should be a summary of errors at the top of a page
+- The focus should be shifted to that summary
+- Each error should link to the input that is invalid
+- Each invalid input should *also* have the error beneath/beside/above it
+
+This means that not only will ATs have clear access to what has happend on the page, but it will be clear for any other user of the product's form.
+
+<<<<<<< HEAD
+`o-forms` implements this behaviour for you if the [error summary options is enabled](#TODO)
```
notlee: 🚓 👮 ~~~~ conflict alert ~~~~ 🚓 👮

ACCESSIBILITY.md
gvonkoss: Many products to exactly that, so I think we do need to point it out

ACCESSIBILITY.md
gvonkoss: No, and it can't, because dyncamically added aria-live attributes don't work

ACCESSIBILITY.md
gvonkoss: wait didn't i fix this 🤔

ACCESSIBILITY.md
gvonkoss: I don't really know... I think the information should be there, but maybe as a comment instead

ACCESSIBILITY.md
notlee: Hmmm if we create the summary dynamically without`aria-live`, how do users add `aria-live`?

ACCESSIBILITY.md
notlee: You did whilst I was reviewing :)

ACCESSIBILITY.md
notlee: Ah okay, maybe a wording tweak?
```suggestion
_Note 2: `aria-hidden` is sometimes used as a CSS selector to visually hide/show elements. However, it's primary functionality is to hide the contents of an element from the AT_
```

ACCESSIBILITY.md
gvonkoss: The summary doesn't need `aria-live`, because the focus shifts onto the summary. `aria-live` will read out changs as they happen, so on `blur`, `input`, `change`, etc

gvonkoss: The users need to add `aria-live` to the markup

notlee: >If you choose not to use the error summary functionality that comes with `o-forms`

I didn't read it properly 😅 thanks!

PR #255, opened 2019-04-16
V7: Add Input/State JS

gvonkoss:
~Extracts `Input` from `Form` to enable state-setting on inputs~

~This is also adding states—my brain feels foggy I can't seem to figure out how best to make the state API public—should it even be under `Forms`? Or under `Input`  😱  OR should it be entirely independent!!!???~

This PR creates an independent `Input` and `State` class, which are both initialised by `Forms`, or can be initialised independently.


src/js/forms.js
```diff
@@ -1,65 +1,51 @@
...
 		if (!this.opts.useBrowserValidation) {
-			this.formEl.setAttribute('novalidate', true);
-			this.formEl.addEventListener('submit', this);
+			this.form.setAttribute('novalidate', true);
+			this.form.addEventListener('submit', this);
 		} else {
-			this.formEl.removeAttribute('novalidate');
-			let submit = this.formEl.querySelector('input[type=submit]');
+			this.form.removeAttribute('novalidate');
+			let submit = this.form.querySelector('[type=submit]');
```
rowanmanning: This should maybe cater for forms with multiple `submit` inputs. I don't know whether we use them anywhere at the moment, but an example use case would be something like GitHub's "Comment" / "Close and comment". Here, the submit that you click changes the behaviour of the form.

src/js/forms.js
```diff
@@ -1,65 +1,51 @@
...
 	*/
 	constructor(formElement, options) {
-		this.formEl = formElement;
-		this.inputFields = Array.from(formElement.querySelectorAll('input', 'select', 'textarea'));
+		if (formElement.nodeName !== 'FORM') {
+			throw new Error(`[data-o-component="o-forms"] must be set on a form element. It is currently set on a '${formElement.nodeName.toLowerCase()}'.`);
+		}
+
+		this.form = formElement;
+		this.formElements = Array.from(this.form.elements, element => new Input(element));
```
rowanmanning: `Array.from(arrayLike, ✨🗺🎉mapFn🎉🗺✨)`

src/js/forms.js
gvonkoss: It does change the behaviour of the form, but only in so far that it validates the whole form before it submits—it has zero opinions as to what happens on the actual submit.

I'll query for multiple submit inputs/buttons to handle cases where there might be more than one 👍

src/js/state.js
```diff
@@ -0,0 +1,77 @@
...
+			this.parent.append(this.stateEl);
+		}
+	}
+
+	/**
+	* Saved state
+	* @access private
+	*/
+	_saved() {
+		this.parent.classList.replace(this.className.saving, this.className.saved);
```
JakeChampion: `classList.replace` does not exist in IE. We would need to add DOMTokenList to the browserFeatures.required array for `classList.replace` to get polyfilled. It would be good to run the tests across our browsers via `obt --browserstack` and ensure they all pass

test/forms.test.js
```diff
@@ -2,133 +2,92 @@
...
-
-		inputFields = document.body.querySelectorAll('input, textarea, select');
-		dateField = inputFields[0];
-		requiredTextField = inputFields[3];
-		optionalTextField = inputFields[4];
-
-		dispatch = (event, element) => element.dispatchEvent(new Event(event));
-		parentClass = (element, validity) => element.closest('.o-forms-input').classList.contains(`o-forms-input--${validity}`);
+	before(() => {
+		parentClass = (element, modifier) => element.closest('.o-forms-input').classList.contains(`o-forms-input--${modifier}`);
```
JakeChampion: Element.closest does not exist in IE. we need to add it to the browserFeatures.required array and then should also run the tests in IE by using `obt --browerstack`.

src/js/forms.js
```diff
@@ -1,65 +1,53 @@
...
-			if (e.target.validity.valid && field.classList.contains(this.className.invalid)) {
-				field.classList.replace(this.className.invalid, this.className.valid);
-			}
-		}
-
 		const RETURN_KEY = 13;
 		if (e.type === 'click' || (e.type === 'keydown' && e.key === RETURN_KEY)) {
-			if (!this.formEl.reportValidity()) {
-				this.inputFields.forEach(input => this.validate(input));
+			if (!this.form.checkValidity()) {
```
JakeChampion: Does `checkValidity` exist in IE? MDN doesn't say yes or no 😔 
image

src/js/forms.js
```diff
@@ -1,65 +1,53 @@
...
+				this.validateForm();
```
notlee: I found this difficult to understand, I read it like "if the form is not valid check the form is valid". Could we make it more clear with comments? It might have been the naming that threw me a little too, `this.form` is an `Element` but `this.formElements` is not an array of `Element`s.

src/js/forms.js
```diff
@@ -68,16 +56,34 @@ class Forms {
...
+			object = {
+				name,
+				element: new State(this.form.elements[name])
+			};
+
+			this.stateArray.push(object);
+		}
+
+		if (!state) {
+			throw new Error(`${state} is not a recognised state, the options are 'saving', 'saved' or 'none'.`);
```
notlee: This error should live in `state.js`, so it throws here and if `state.js` is used directly.

src/js/state.js
```diff
@@ -0,0 +1,77 @@
...
+	* Class constructor.
+	* @param {RadioNodeList} [inputs] - A NodeList of radio input elements
+	*/
+	constructor(inputs) {
+		this.inputs = inputs;
+		this.parent = inputs.length > 1 ? this.inputs[0].closest('.o-forms-input') : null;
+
+		this._verify();
+
+		this.stateEl = null;
```
notlee: It's maybe more intuitive to set this to the element immediately in the constructor? Especially as a public property.

src/js/forms.js
gvonkoss: it seems to, as far as IE10 https://caniuse.com/#search=reportV. Perhaps forms should default to using browser validation for <IE10

src/js/forms.js
```diff
@@ -1,65 +1,53 @@
...
 	constructor(formElement, options) {
-		this.formEl = formElement;
-		this.inputFields = Array.from(formElement.querySelectorAll('input', 'select', 'textarea'));
+		if (formElement.nodeName !== 'FORM') {
+			throw new Error(`[data-o-component="o-forms"] must be set on a form element. It is currently set on a '${formElement.nodeName.toLowerCase()}'.`);
+		}
+
+		this.form = formElement;
+		this.formElements = Array.from(this.form.elements, element => new Input(element));
+		this.stateArray = [];
```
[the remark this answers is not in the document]
gvonkoss: It needs to be available to the whole form in order to be able to call `form.setState(...)`

demos/src/demo.js
```diff
@@ -4,7 +4,6 @@ document.addEventListener("DOMContentLoaded", function () {
 	document.dispatchEvent(new CustomEvent('o.DOMContentLoaded'));
 });
 
-// to demonstrate :invalid pseudo-states work (without user interaction which is how they should work in prod)
```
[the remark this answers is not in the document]
gvonkoss: I wanted to avoid adding excessive data points to the form JSONs because they are already getting huge, and if they were set to a default in the templates, everything would be invalid.

src/js/forms.js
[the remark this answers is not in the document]
notlee: It can still be used in `form.setState`? A private property can be used anywhere within the object.

demos/src/demo.js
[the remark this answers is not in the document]
notlee: I see what you mean, it might be better to do that though as the markup will be incorrect for users in the registry for the core experience of when they hit the `html` tab?

src/js/forms.js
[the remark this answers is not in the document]
gvonkoss: it borked when I tried it 🤷‍♀️ maybe I did it wrong?

PR #257, opened 2019-04-29
Switch require to import in the docs

rowanmanning:
None of the examples in the docs currently work 🙈


j-mes:
Fantastic to see that we highlighted something 😄

PR #259, opened 2019-05-01
V7: Documentation

gvonkoss:
Documentation for v7!

there are some additions to the demo because of the changes that happened in the `a11y` branch.


README.md
```diff
@@ -2,622 +2,566 @@
...
-
-For more details on specific mixins [browse the SassDoc documentation of the module](http://sassdoc.webservices.ft.com/v1/sassdoc/o-forms/#o-forms).
-
-### JavaScript
 
-`o-forms` provides some JavaScript to help with validating form inputs. The validation works with the browsers built-in validation API to add the appropriate `o-forms` classes when a user is completing or submitting a form.
+### Errors
 
-By default, `o-forms` listens to the `blur` event of an input field, when a user leaves a field, the input will be validated and if an invalid input is found, the error class will be added to the input.
+In trms of invalid inputs, `o-forms` has a build in mechanism to display custom errors and an error summary. If neither of these work for your product, you can also default to [native browser validation](#form-instance).
```
JakeChampion:
```suggestion
In terms of invalid inputs, `o-forms` has a build in mechanism to display custom errors and an error summary. If neither of these work for your product, you can also default to [native browser validation](#form-instance).
```

README.md
```diff
@@ -2,622 +2,566 @@
...
+#### Error Summary
+`o-forms` also generates an error message element when a form is submitted and invalid inputs are recognised.
 
-By default `o-forms` tests inputs on the `blur` event for each input. To defer the validation to the `submit` event of the form, you can pass an config object to set the `testEvent` when constructing `o-forms`:
+This feature will collect the custom messages of the invalid fields if they are present in the markup, or will default to the browsers native error message if they aren't. It will associate the message to the title of the input that is invalid, generate a list of links at the top of the form, and focus on the first link.
 
-```js
-const OForms = require('o-forms');
-const formsEl = document.querySelector('#form-element');
+This markup will always generated dynamically if the [errorSummary option](#form-instance) is set to `true`.
```
JakeChampion:
```suggestion
This markup will always be generated dynamically if the [errorSummary option](#form-instance) is set to `true`.
```

README.md
```diff
@@ -2,622 +2,566 @@
...
 
-const forms = new OForms(formsEl, { testEvent: 'submit' });
+## Sass
+You can include styles for every input type and feature by calling the `o-forms` primary mixin:
```
JakeChampion: I may be misunderstanding, but I think the primary mixin is called `oForms` and the sass file to import is called `o-forms`. 
I'm not sure if my suggestion is correct though so feel free to resolve this without changing it.
```suggestion
You can include styles for every input type and feature by calling the `oForms` primary mixin:
```

README.md
```diff
@@ -2,622 +2,566 @@
...
+	- `'inline'`
+	- `'right'`
+	- `'inverse'`
+	- `'small'`
+	- `'state'`
+	- `'suffix'`
+
+_Note 1: `valid` and `invalid` styles are output with the base of `o-forms` so there is no need to include them in the list above._
+
+_Note 2: It's important to remember that not all features will apply to all input types. If there are any features you would like to see added to an input, please [get in touch with the team](#contact)._
```
JakeChampion: 👍

README.md
```diff
@@ -2,622 +2,566 @@
...
+- `'features'`:
+	- `'disabled'`
+	- `'inline'`
+	- `'right'`
+	- `'inverse'`
+	- `'small'`
+	- `'state'`
+	- `'suffix'`
+
+_Note 1: `valid` and `invalid` styles are output with the base of `o-forms` so there is no need to include them in the list above._
```
JakeChampion: 👍

MIGRATION.md
```diff
@@ -0,0 +1,183 @@
...
+
+Version 7 introduces a complete redesign to the `o-forms` markup and API, and makes small design changes and additions.
+
+All Sass mixins have been removed, and have been replaced with two public mixins:
+- `oForms()`
+- `oFormsAddCustom()`
+
+The markup has been changed entirely to accomodate the following structure:
+```
+┌— field container (.o-forms-field) —————┐
```
JakeChampion: 🙌

README.md
gvonkoss: Oh it was meant to imply the primary mixin that belongs to `o-forms` but yeah now that you point it out it is kind of confusing 🤔

gvonkoss:
At the moment we seem to have a disagreement about the naming convention for one of forms' outliers—namely anchors. I'm jotting this down here because we have had this conversation face to face and not written anything down anywhere.

The  anchors are made to look like a radio box,  and they've received the modifier class 'pseudo-radio-box', but have a few different classes for styling, as well (`o-forms__anchor`, `o-forms__anchor--current`, etc). 

Along with the box-style radio buttons, they are the only elements that can be customised.
I think they still apply as elements or input types, given their unusual situation, so I decided to refer to them as 'anchor' in the primary mixin and in the custom mixin, but @notlee  argues that it isn't an intuitive naming system, and prefers 'pseudo-radio-box':
https://github.com/Financial-Times/o-forms/pull/259#discussion_r280745062
https://github.com/Financial-Times/o-forms/pull/259#discussion_r280745278
https://github.com/Financial-Times/o-forms/pull/259#discussion_r280746750

If we could get all of our opinions on this and break the stalemate, we can release o-forms!

README.md
```diff
@@ -2,622 +2,558 @@
...
 ```
+[_See the full markup for a box-style radio button in the registry_](https://registry.origami.ft.com/components/o-forms#radio-box)
 
-[Radio control examples](https://www.ft.com/__origami/service/build/v2/demos/o-forms/radios)
-[Checkbox examples](https://www.ft.com/__origami/service/build/v2/demos/o-forms/checkboxes)
-
-#### Styled radio buttons
-
-To produce styled radio buttons wrap radio inputs with  `.o-forms__group .o-forms__group--inline-together` and use the class `.o-forms__radio-button` on the radio input.
+#### `input[type=radio]`
```
rowanmanning: Looking at naming as well, but just spotted this
```suggestion
#### `input[type=checkbox]`
```

JakeChampion:
I'd go with both (link/anchor/pseudo-radio-box) because a new designer/developer looking to use o-forms will likely search for anchor or link etc and a developer/designer who has a current o-forms component which uses pseudo-radio-box can still find the documentation by searching for pseudo-radio-box.

For any mixin/function input, I'd go with either just `anchor` because `anchor` is the element in `o-forms__anchor` or I'd go with both `anchor` and `pseudo-radio-box`.

rowanmanning:
I agree that `o-forms__anchor` isn't intuitive, I think it's mostly the fact that it's named after the element rather than what the element _does_. The class is kind of named for an implementation detail, and if (for example) we later wanted a `button` element to be valid in this case then the class no longer makes sense. I'm trying to think of an alternative, but so far `pseudo-radio-box` makes more sense to me.

Agree with Jake that the documentation should include both `anchor` and whatever we go with for easy finding of this.

gvonkoss:
So we're discussing a number of things, because we can refer to the anchor as both in the documentation but the code is currently only handling one of those two things—and we can change the naming for the classes, but that would be something like... `o-forms-input--pseudo-radio-box__??`

rowanmanning:
OK, what if we call the feature `radio-box-link`? It indicates:

a) what the thing looks like visually and
b) how it behaves – it's a link and will update the URL

Would mean class name changes as well, though. Bear with me here, if the thing can _only_ ever contain links, and links can't be nested, and there's one use-case is it OK to style the `a` element without a class? The only issue I have is what to call the `current` class... Something like:

```html
<div class="o-forms-input o-forms-input--radio-box-link">
	<a class="<SOME KIND OF CURRENT CLASS>" href="#">Anchor Link A</a>
	<a href="#">Anchor Link B</a>
</div>
```

Or am I just further muddying the waters?

gvonkoss:
All inputs are actually styled by nesting the element in under the class modifier so I think that could work 👍 

And it really shouldn't ever be anything other than a link because then we would invite more outliers to be added

JakeChampion:
I like @rowanmanning's idea, let's do that :-)

PR #274, opened 2019-06-12
Backport new border colours to v6 from v7.

notlee:
This improves the contrast ratio against paper, improving
accessibility whilst we work on a v7 rollout.


notlee:
Released in v6.0.3

PR #278, opened 2019-07-02
Introduce o-spacing.

notlee:
o-spacing functions replace:
- oTypographyMargin
- oTypographyPadding
- oTypographySpacingSize

Although this introduced a new dependency, o-spacing is new and should not be considered breaking.

### DIFF
--- a/src/scss/_pseudo-radio-link.scss
+++ b/src/scss/_pseudo-radio-link.scss
@@ -4,9 +4,8 @@
 @mixin _oFormsPseudoRadioLink {
 	.o-forms-input--pseudo-radio-link {
 		@include _oFormsControlsBoxContainer();
-		@include oTypographyMargin($top: 0);
-
 		@include _oFormsPseudoRadioLinkStyles();
+		margin-top: 0;
 	}
 }
 
--- a/src/scss/shared/_control-inputs.scss
+++ b/src/scss/shared/_control-inputs.scss
@@ -42,7 +42,7 @@
 	float: left;
 	margin: 0;
 	min-height: $_o-forms-spacing-seven;
-	min-width: oTypographySpacingSize(15);
+	min-width: 8ch;
 }
 
 /// @access private
--- a/bower.json
+++ b/bower.json
@@ -1,16 +1,17 @@
 {
-	"name": "o-forms",
-	"main": [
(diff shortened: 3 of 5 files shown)


src/scss/_pseudo-radio-link.scss
```diff
@@ -4,9 +4,8 @@
 @mixin _oFormsPseudoRadioLink {
 	.o-forms-input--pseudo-radio-link {
 		@include _oFormsControlsBoxContainer();
-		@include oTypographyMargin($top: 0);
-
 		@include _oFormsPseudoRadioLinkStyles();
+		margin-top: 0;
```
chee: are these the same?!

src/scss/shared/_control-inputs.scss
```diff
@@ -42,7 +42,7 @@
 	float: left;
 	margin: 0;
 	min-height: $_o-forms-spacing-seven;
-	min-width: oTypographySpacingSize(15);
+	min-width: 8ch;
```
chee: is this width not to be based on what the spacing widths are based on?

src/scss/_pseudo-radio-link.scss
notlee: Yup! [oTypographyMargin](https://github.com/Financial-Times/o-typography/blob/1725d765269f32cb0b077004e44c7df1e20da9a7/src/scss/_mixins.scss#L37) outputs the long form syntax for top and or bottom margins. With just one `$top` argument it just outputs the top margin. 

Also `_oFormsPseudoRadioLinkStyles` doesn't output a margin property, so the re-order is ok.

src/scss/shared/_control-inputs.scss
notlee: It's less typical to use them for widths, usually that's left to the content length or a containing element. In this case ([box style radio inputs](https://registry.origami.ft.com/components/o-forms@7.0.9#demo-box-styled-radio-inputs))  most of the time the length of the content in the box will set the width, so I switched the min-width out to also be derived from the width of the content.

src/scss/_pseudo-radio-link.scss
chee: !!!

PR #290, opened 2019-09-27
Let's use origami-ci!

chee:


origamiserviceuser:
o-forms bundle size difference from 7.1.0 to 51f15a791717172748053eeb68496f02724b134f
No bundle size difference found.

PR #294, opened 2019-11-15
Upgrade dependencies to the latest beta

notlee:
**ready for review but do not merge until whitelabel crimson and jade colours are restored**


origamiserviceuser:
o-forms bundle size difference from 7.1.1 to 6a093e0bc5c6d2c10ffe25704f25ca4bab89ed2d
js: 3.61kb increase (1.27kb/gzip)
css, master: 1.04kb increase (0.10kb/gzip)
css, internal: 1.04kb increase (0.10kb/gzip)

PR #296, opened 2019-11-22
Fix the demos in non-master brands

rowanmanning:


origamiserviceuser:
o-forms bundle size difference from 7.1.1 to aaee7448d068e1ddb024c57555aa73e0fee7c198

PR #297, opened 2019-11-22
Restore Crimson and Jade

rowanmanning:


origamiserviceuser:
o-forms bundle size difference from 7.1.1 to 16a6ed78567897505a67683710d2644592b253a9

PR #298, opened 2019-11-22
Fix buttons in the demos

rowanmanning:
They weren't using the secondary modifier.


origamiserviceuser:
o-forms bundle size difference from 7.1.1 to 83b6a95754f201a3a1be54ffebc58bd7287f1c58

PR #299, opened 2019-11-22
Add v8 to the migration guide

rowanmanning:


origamiserviceuser:
o-forms bundle size difference from 7.1.1 to f7172c184294fe2eea2a3a6a7da8541e550df686

PR #301, opened 2019-12-04
Fix checked and disabled styles on checkboxes with extra content.

dapenguin:
## Problems
When using a checkbox that has extra content (i.e. more than one `<span class="o-forms-input__label">`), there are two issues:

1) Clicking on either the label or the checkbox does not visibly check the checkbox. The field will be marked as checked in the DOM, but it just won't be visually apparent. This issue can be seen on the [Checkboxes demo for `o-forms` version `8.0.0`](https://registry.origami.ft.com/components/o-forms@8.0.0#demo-checkboxes) under "Checkboxes with extra content."
2) When disabled, only the first label for the checkbox would have the disabled styling applied. Also, the checkbox would not have the disabled styling.

I've only checked in Chrome and Firefox, but the issue occurs in both.

This is happening because each bit of content in the label has its own checkbox displayed via the psuedo elements, and they all sit on top of one another. By using an _adjacent_ sibling selector (`+`), the checked and disabled styling is only applied to the psuedo element for the first bit of content in the label, which is being hidden by the psuedo elements for the other bits of content.

## Solution
To fix this issue, I've used a _general_ sibling selector (`~`) instead so that all content in the label has the styling applied, rather than just the first bit of content.

I've also added an additional checkbox to the demo to make sure that the disabled styling is applied when there is extra content in the label. I'm not precious about that staying in there though, just thought it may be useful.

## Screenshots

**Before**
checkboxes-extra-content-before

**After**
checkboxes-extra-content-after

### DIFF
--- a/demos/src/data/checkboxes.json
+++ b/demos/src/data/checkboxes.json
@@ -226,6 +226,13 @@
 				{
 					"type": "checkbox",
 					"name": "extra",
+					"value": "This is a disabled checkbox with extra content.",
+					"additionalContent": "Here is the extra content for this checkbox.",
+					"disabled": true
+				},
+				{
+					"type": "checkbox",
+					"name": "extra",
 					"value": "By submitting this form, you indicate your consent to also being contacted by the Financial Times by email. post or phone about our other products, services or special offers unless you untick this box. You can change your preferences anytime in the My Account section.",
 					"checked": true
 				}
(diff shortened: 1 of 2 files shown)


origamiserviceuser:
o-forms bundle size difference from 8.1.0 to abe89a9e725f53480057413cc5f1f2614a55f208

origamiserviceuser:
o-forms bundle size difference from 8.1.0 to fa02e7146d075b5752216eafcf1f606c94391804

PR #302, opened 2019-12-11
Output custom form icon base styles only once.

notlee:
Also make the `o-forms-input__link--current` more specific.

Source order seems to be behind the icon appearing white on myft:
70637531-4a12ad00-1c2f-11ea-85b5-a812bb15beac

### DIFF
--- a/src/scss/modifiers/_custom.scss
+++ b/src/scss/modifiers/_custom.scss
@@ -42,7 +42,7 @@
 		min-width: unset;
 	}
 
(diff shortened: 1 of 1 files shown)


origamiserviceuser:
o-forms bundle size difference from 8.1.1 to 93126d0c0fab6cc2b34c329e3dfa7a9013bde69b

notlee:
The team are away until the new year, so I'm going to take a little liberty and merge this 🙃

Issue #304, opened 2020-01-02
Saving indicator doesn't stay attached to inline box radios


edds:
## What

This is a screenshot of the examples where the saving is nicely ‘attached’ to the bottom of the radio buttons.
![Screenshot 2019-12-23 at 16 09 56](https://user-images.githubusercontent.com/35035/71660662-ae62f100-2d43-11ea-9051-c0ca68376092.png)

But if the description on the left hand side gets really long (for example if it’s explaining the consequences of un-checking the radio. Then the saving spinner gets shunted way do the page, and it no longer looks attached to the thing you’ve just changed.
![Screenshot 2019-12-23 at 16 09 35](https://user-images.githubusercontent.com/35035/71660670-b6229580-2d43-11ea-8b1c-d2bd06d21d12.png)

## Details
This is due to the saving indicator having a position absolute relative to the bottom of the container. And as we've switched to flex-box in the most recent o-forms major this is now the same height as the label to the left of it. You can see the desired behaviour by looking at v7 version of o-forms

notlee:
yikes, thanks for the detailed issue

notlee:
This should be fixed in v8.1.4

PR #306, opened 2020-01-07
Position state and error elements against inputs when labels are long.

notlee:
When a label and input are inline, with label to the left and
input to the right, the state/error element is positioned on
the right under the input. If the label is long the status/error
message is pushed down away from the input due to absolute
positioning against their shared container.

71660670-b6229580-2d43-11ea-8b1c-d2bd06d21d12

Absolute positioning is used with a negative bottom value so the
state does not push content down the page when added dynamically.
The state sits within the fields margin instead. We can use
relative positioning and a negative margin instead. This keeps
the state close to the input when the label becomes long, but also
allows the state to avoid pushing other page content down by
negating the containers margin.

Screenshot 2020-01-07 at 13 29 54
Screenshot 2020-01-07 at 13 30 38

Relates to: #304

(Also fix "vertical centered inline radio box" demo.)

### DIFF
--- a/demos/src/data/radio-box.json
+++ b/demos/src/data/radio-box.json
@@ -174,14 +174,14 @@
 			"field": {
 				"type": "radio-box",
 				"aria": {
-					"label": "saving-state-group-title",
-					"info": "saving-state-group-info"
+					"label": "saving-state-group-title-long",
+					"info": "saving-state-group-info-long"
 				},
 				"modifiers": ["saving"],
 				"state": "Saving"
 			},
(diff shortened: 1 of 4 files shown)


origamiserviceuser:
o-forms bundle size difference from 8.1.3 to 470cc09da82e8b7b688bb61c9060fee43894cf1f

origamiserviceuser:
o-forms bundle size difference from 8.1.3 to beeb58afbeac4de52c9f7bc7196b203ae3868318

Issue #285, opened 2019-09-02
Checkbox pseudo element positions are not calculated correctly


bbcreatv:
## What

[Bug]: On mobile widths the checkbox pseudo element positions are not calculated correctly - with no modifications to css. On the origami demo page it seems to be happening to the right side checkbox only.

![Screenshot 2019-09-02_22-40-58-161](https://user-images.githubusercontent.com/3500928/64132830-c563f780-cdd2-11e9-9109-27dbdf2259d7.png)

## Details
- Chrome 76
- iPhone X/Desktop
- o-forms v7.0.11
![Screenshot 2019-09-02_22-31-48-886](https://user-images.githubusercontent.com/3500928/64132732-2808c380-cdd2-11e9-96d6-238257b4e0b7.png)

notlee:
Looks like we missed this issue! Thanks for making it detailed 😊 
I've opened a PR with a fix: https://github.com/Financial-Times/o-forms/pull/307

notlee:
A fix was released in v8.1.5

PR #307, opened 2020-01-07
Hide checkbox inputs.

notlee:
We recreate checkboxes with pseudo elements so hide the original
inputs visually.

Relates to https://github.com/Financial-Times/o-forms/issues/285


origamiserviceuser:
o-forms bundle size difference from 8.1.4 to cce64bd649476d6936acd9bd215898ff9d9f80f1
css, master: 0.36kb increase (0.01kb/gzip)
css, internal: 0.36kb increase (0.01kb/gzip)
css, whitelabel: 0.32kb increase (0.01kb/gzip)
An insignificant difference was also found for: js

PR #308, opened 2020-01-10
Test one form does not validate on submit of a different form.

notlee:
Turns out this is not a problem,
I thought it might have been,
but now we have tests to confirm it.


origamiserviceuser:
o-forms bundle size difference from 8.1.5 to edd1905bbd3cfcdad0c90dbade8ca7abb3c68565

PR #309, opened 2020-01-10
Show invalid inputs with missing ids in the error summary.

notlee:
Warn in the console that the error summary cannot link to an input
with no id. This is better than not showing the user the error
in the summary at all. I considered generating an id but thought
it may risk collisions, or users attempting to hang logic from
dynamic ids which may change.

Also warn in the console if the error summary is enabled but an
invalid input does not have a field title to show in the error
summary. For example, using the label of a single checkbox
"I accept these terms." would not make sense in an error summary;
the developer should add a field title like "terms and conditions"
if they would also like to use the error summary with such an input.

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,11 +456,11 @@ In order to provide customised error messages for an invalid input field, you'll
 The message is hidden by default, until the input field becomes invalid.
 
 #### Error Summary
-`o-forms` also generates an error message element when a form is submitted and invalid inputs are recognised.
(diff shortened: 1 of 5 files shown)


origamiserviceuser:
o-forms bundle size difference from 8.1.5 to 83348c2d2f011daa30df1adfe95f4c836a8c0c1c
js: 0.65kb increase (0.27kb/gzip)
An insignificant difference was also found for: css (master), css (internal), css (whitelabel)

PR #310, opened 2020-01-13
fix: update form.summary with new element

alexnaish:
Upon re-submitting a form with invalid elements that is already showing a summary, the replacement of the previous summary wasn't updating the DOM reference. This lead to an error when attempting to replace the summary as the stored reference was no longer within the DOM tree. This meant we would receive:

```
Uncaught DOMException: Failed to execute 'replaceChild' on 'Node': The node to be replaced is not a child of this node.
```

### DIFF
--- a/src/js/forms.js
+++ b/src/js/forms.js
@@ -85,7 +85,9 @@ class Forms {
 			if (checkedElements.some(input => input.valid === false)) {
 				if (this.opts.errorSummary) {
 					if (this.summary) {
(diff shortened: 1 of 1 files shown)


origamiserviceuser:
o-forms bundle size difference from 8.1.6 to a5653d511d09118a5323dc05fefca7edd23d4c35

JakeChampion:
We should look to add a test for this to ensure we do not regress on this bug fix

PR #312, opened 2020-01-27
Update support channel to origami-support

chee:


origamiserviceuser:
o-forms bundle size difference from 8.1.8 to 9b978bbb65fb133261ad38165fe419ff810f222e

PR #313, opened 2020-01-27
Remove `classList.replace` for IE11 support.

notlee:


origamiserviceuser:
o-forms bundle size difference from 8.1.8 to 98d1a648f57bc449845cecc5c87b0b695540ea4e

PR #314, opened 2020-01-27
Document other required polyfills.

notlee:
These polyfills are already required by o-forms (so this isn't
a breaking change). As our users are already including these
polyfills I don't think we have to release a patch that removes
these features -- but they should be documented for the future.


origamiserviceuser:
o-forms bundle size difference from 8.1.9 to 8a4580a872e37aee0b35a615bf87423385a731f1

Issue #318, opened 2020-02-12
IE11 select arrow positioning


bbcreatv:
## What

Possibly this a known issue but on IE11 the dropdown arrows positioning is off on wide selects
## Details

- IE11 (Windows 10)
- o-forms 8.1.1

![image](https://user-images.githubusercontent.com/3500928/74329728-9fcf1700-4d90-11ea-9913-c8288ab0e8eb.png)

PR #319, opened 2020-02-13
Fix select input arrow position in IE11

notlee:
>If you only set one dimension using background-size and the other
>to auto, IE will choose the aspect ratio of the element rather
>than the intrinsic dimensions of the SVG graphic.

https://thatemil.com/blog/2015/03/15/sizing-svg-background-images-in-internet-explorer/

Fixes: https://github.com/Financial-Times/o-forms/issues/318

Screenshot 2020-02-13 at 17 40 15


origamiserviceuser:
o-forms bundle size difference from 8.1.11 to 15f3c7bf7da43729039f3b65dfe9a136b8a00fa4

chee:
:+1:

PR #323, opened 2020-02-27
Remove default browser styles.

notlee:
E.g. we do not want a drop shadow or border radius in mobile Safari

Before 😱:
Screenshot 2020-02-27 at 15 00 29

After:
Screenshot 2020-02-27 at 15 01 18


origamiserviceuser:
o-forms bundle size difference from 8.1.13 to 218f5e35a240cea605a8973fd754837788f544a1

PR #325, opened 2020-02-27
Correct toggle spacing so the label is not obscured.

notlee:
Default browser styles were removed in the last commit but
o-forms toggles relied on hidden checkbox dimensions for
layout.

before/after
Screenshot 2020-02-27 at 16 34 29
Screenshot 2020-02-27 at 16 39 04


origamiserviceuser:
o-forms bundle size difference from 8.2.0 to 70c250a54886db22ef24b709945078a131bc00e4

PR #327, opened 2020-03-09
test branch-package-tests branch of oci

chee:


origamiserviceuser:
o-forms bundle size difference from 8.2.1 to 108aac6398cc92c6d562be2c05f066db140a8188

chee:
ok so this works

PR #331, opened 2020-03-26
Allow a declaritive error summary.

notlee:
For users who do not use o-forms JavaScript.


origamiserviceuser:
o-forms bundle size difference from 8.2.2 to a819269b786156e46dddcef3c6ae005021afebb8
js: 0.31kb increase (0.10kb/gzip)
An insignificant difference was also found for: css (master), css (internal), css (whitelabel)

PR #332, opened 2020-04-01
Ensure Safari uses the correct text colour for disabled inputs.

notlee:
before/after
Screenshot 2020-04-01 at 10 36 57


origamiserviceuser:
o-forms bundle size difference from 8.3.0 to 14a825b88e9653ec350f5ec4a4504c9204dd266c

PR #334, opened 2020-04-01
Stop using magic webpack imports

JakeChampion:
Stop using the magic imports for proclaim and sinon and pull them from the global because that's where Karma has added them. This is required for obt v10 to be released.


origamiserviceuser:
o-forms bundle size difference from 8.3.1 to 620800bcf70c41afa895ab595be63f648f6f9282

PR #352, opened 2020-06-24
Hide native radios and checkboxes while leaving them accessible to dragons

chee:
this should fix these:
https://financialtimes.atlassian.net/browse/NOPS-235
https://financialtimes.atlassian.net/secure/RapidBoard.jspa?rapidView=1108&projectKey=ACC&modal=detail&selectedIssue=ACC-423 

and this https://github.com/Financial-Times/n-feedback/pull/88 

tested in:

Dragon on IE11 on Windows 10
Firefox on Linux

hopefully a kind robot can show us some visual regression tests from other browsers so we know we haven't made anything worse?

Also worth testing it has the same screenreader experience as the previous version

```
npm i -g origami-build-tools
git clone git@github.com:financial-times/o-forms
cd o-forms
obt install
obt dev
```

and try out http://localhost:8999/checkboxes and http://localhost:8999/radio-round

PR #353, opened 2020-06-25
Also fix the inputs on the radio boxes

chee:
following on from this: https://github.com/Financial-Times/o-forms/pull/352

@taraojo has pointed out that the `_oFormsControlsBase` mixin isn't used in the box-styled radio inputs!

i've added it here directly, but maybe we should also open an issue about getting the box-styled ones to actually use that mixin? because as @taraojo also points out, the comment on the mixin actually mentions box-style radios

### DIFF
--- a/src/scss/_radio-box.scss
+++ b/src/scss/_radio-box.scss
@@ -36,7 +36,13 @@
 
 		input[type=radio] { // stylelint-disable-line selector-no-qualifying-type
 			position: absolute;
+			// Hide visually while remaining accessible to voice control like Dragon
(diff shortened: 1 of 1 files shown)


taraojo:
> maybe we should also open an issue about getting the box-styled ones to actually use that mixin?

Was this because `_oFormsControlsBase` doesn't have could be applied to the box-style radio button? I saw there is also `_oFormsControlsBoxBase` that is used for box-style radios. And "anchor controls", though I'm not sure what they are 😅

origamiserviceuser:
:tada: This PR is included in version v8.3.10 :tada:

PR #356, opened 2020-08-13
Use the new $default-branch macro instead of hardcoding the branch name

JakeChampion:
This will help with any future work on renaming the default branch name.


origamiserviceuser:
o-forms bundle size difference from 8.3.11 to 6f58a5ce8735c94d8f50e39e1543d407b1f8d351

PR #362, opened 2020-09-04
add extension to relative imports to make the code work directly in a browser

JakeChampion:
relative imports require the full path including the extension because browsers use the import name to make a network request and they do not implicitly add a .js file extension if one is missing


origamiserviceuser:
o-forms bundle size difference from 8.3.11 to 29a4bd25fdc747cf77eeb24a8ebbdbe6903ab654

origamiserviceuser:
:tada: This PR is included in version v8.3.12 :tada:

Issue #303, opened 2019-12-12
correct suffix colour with valid input


notlee:
`o-forms-input--valid` styles all children not just inputs.

This means a suffixed button will turn green on a valid field.

(top: normal, bottom: valid class applied which makes the button green. only the input colour should be effected)
<img width="545" alt="Screenshot 2019-12-12 at 14 12 53" src="https://user-images.githubusercontent.com/10405691/70719392-d1245b80-1ce9-11ea-83ba-79db2a8083fa.png">

https://github.com/Financial-Times/o-forms/blob/5771c0d00241c0aeab953b2f98e1a683527f499e/src/scss/shared/_validity.scss#L5

PR #366, opened 2020-09-24
Do not style suffix buttons based on validity.

notlee:
Fixes: https://github.com/Financial-Times/o-forms/issues/303

[`:valid`](https://developer.mozilla.org/en-US/docs/Web/CSS/:valid) is targetting `button` since it is a form element.
Ideally I think we'd use a class here BEM style rather than exclude specific elements.

### DIFF
--- a/demos/src/interactive-suffix.mustache
+++ b/demos/src/interactive-suffix.mustache
@@ -1,0 +1,12 @@
+<form action="" data-o-component="o-forms">
+	<label class="o-forms-field">
+		<span class="o-forms-title">
+			<span class="o-forms-title__main">Text input with suffix</span>
+		</span>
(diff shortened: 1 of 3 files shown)


origamiserviceuser:
o-forms bundle size difference from 8.3.12 to b5d8db25f7b26105cfdfdb65323927f4c6a1fcf3

JakeChampion (approved):

origamiserviceuser:
:tada: This PR is included in version v8.3.13 :tada:

origamiserviceuser:
o-forms bundle size difference from 8.3.12 to 97a731fdbb9a5497717537f67e1b12753d0d0ef1

PR #374, opened 2020-11-23
Make getDataAttributes ignore data attributes which are not in the component's namespace

JakeChampion:
This keeps the method aligned with the definition in the component template https://github.com/Financial-Times/create-origami-component/pull/214


origamiserviceuser:
o-forms bundle size difference from 8.3.14 to cb121379e77531c926b385e760dcca0662bbbfe3

PR #377, opened 2020-12-04
Add better error styles for whitelabel brand in inverse mode

JakeChampion:
Currently the demo for inverse fails for whitelabel as it is using `slate` colour which does not exist in the whitelabel colour palette. I've switched to a hardcoded value of `slate` to make it work.

This also updates the error states in whitelabel's inverse mode to match the new design used in internal and master brand

image

### DIFF
--- a/demos/src/inverse.scss
+++ b/demos/src/inverse.scss
@@ -1,5 +1,7 @@
 @import "./demo";
 
 body {
-	background: oColorsByName('slate');
+	// This is the slate colour in master and internal
+	// We hardcode the value because the whitelabel palette does not contain slate.
+	background: #262a33;
 }
--- a/demos/src/inverse.scss
+++ b/demos/src/inverse.scss
@@ -1,7 +1,10 @@
 @import "./demo";
 
 body {
-	// This is the slate colour in master and internal
-	// We hardcode the value because the whitelabel palette does not contain slate.
-	background: #262a33;
+	// slate is in the master and internal brand palette
+	background: if(
+		index(('master', 'internal'), oBrandGetCurrentBrand()) != null,
+		oColorsByName('slate'),
+		oColorsByName('black')
+	);
 }
--- a/origami.json
+++ b/origami.json
@@ -40,7 +40,16 @@
 			"title": "Inverse Form",
 			"template": "/demos/src/inverse-form.mustache",
 			"description": "Inverse form",
-			"sass": "demos/src/inverse.scss"
+			"sass": "demos/src/inverse.scss",
+			"brands": ["master", "internal"]
+		},
+		{
+			"name": "inverse-form-whitelabel",
+			"title": "Inverse Form",
+			"template": "/demos/src/inverse-form-whitelabel.mustache",
+			"description": "Inverse form",
+			"sass": "demos/src/inverse.scss",
+			"brands": ["master", "internal", "whitelabel"]
 		},
 		{
 			"name": "pseudo-radio-links",
--- a/origami.json
+++ b/origami.json
@@ -49,7 +49,7 @@
 			"template": "/demos/src/inverse-form-whitelabel.mustache",
 			"description": "Inverse form",
 			"sass": "demos/src/inverse.scss",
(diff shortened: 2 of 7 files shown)


origamiserviceuser:
o-forms bundle size difference from 8.4.0 to 710b9fc00b8b7612374da3dcec295d1f303a96d5
css, whitelabel: 0.41kb increase (0.04kb/gzip)
An insignificant difference was also found for: js, css (master), css (internal)

notlee:
looks like the button "let's get started" doesn't have much contrast here -- maybe we need a whitelabel specific demo that uses an inverse button and black background?

JakeChampion:
> looks like the button "let's get started" doesn't have much contrast here -- maybe we need a whitelabel specific demo that uses an inverse button and black background?

I've done that now 👍 

image

src/js/error-summary.js
```diff
@@ -17,6 +17,12 @@ class ErrorSummary {
 	*/
 	constructor(elements) {
 		this.elements = elements;
+		const hasAnInverseField = elements.some(elem => {
+			if (elem.field) {
+				return elem.field.classList.contains('o-forms-field--inverse');
+			}
+		});
+		this.inverse = hasAnInverseField;
```
JakeChampion, line 25: I missed this in the original implementation. We need to ensure that the error summary created by this class uses the correct design. I.E. an inverse design for an inverse form.

The way I've done that is to check if any form field has an inverse modifier class and if any do, then the error summary will be inverse style.

origamiserviceuser:
o-forms bundle size difference from 8.4.0 to b4cd1bbdabf907ed3c78b637be365e62d93379c8
css, whitelabel: 0.48kb increase (0.04kb/gzip)
An insignificant difference was also found for: js, css (master), css (internal)

origami.json
```diff
@@ -40,7 +40,16 @@
...
+			"sass": "demos/src/inverse.scss",
+			"brands": ["master", "internal"]
+		},
+		{
+			"name": "inverse-form-whitelabel",
+			"title": "Inverse Form",
+			"template": "/demos/src/inverse-form-whitelabel.mustache",
+			"description": "Inverse form",
+			"sass": "demos/src/inverse.scss",
+			"brands": ["master", "internal", "whitelabel"]
```
notlee, line 52:
```suggestion
			"brands": ["whitelabel"]
```

origamiserviceuser:
o-forms bundle size difference from 8.4.0 to 70f2a04fa189ad537eac83480418b7d0115d898c
css, whitelabel: 0.48kb increase (0.04kb/gzip)
An insignificant difference was also found for: js, css (master), css (internal)

demos/src/inverse.scss
```diff
@@ -1,5 +1,7 @@
 @import "./demo";
 
 body {
-	background: oColorsByName('slate');
+	// This is the slate colour in master and internal
+	// We hardcode the value because the whitelabel palette does not contain slate.
+	background: #262a33;
```
notlee, line 6:
```suggestion
	// slate is in the master and internal brand palette
	background: if(
		index(('master', 'internal'), oBrandGetCurrentBrand()) != null,
		oColorsByName('slate'),
		oColorsByName('black')
	);
```

How about this?

notlee: If you think that looks good it would apply to the `.demo-inverse` class too

origamiserviceuser:
o-forms bundle size difference from 8.4.0 to 2ed6cb8d942c2a26ea667b7709a497d1f768eec6
css, whitelabel: 0.48kb increase (0.04kb/gzip)
An insignificant difference was also found for: js, css (master), css (internal)

origamiserviceuser:
o-forms bundle size difference from 8.4.0 to 4fb8b7aeeb21e25eb3b1ee6caa41c74849003dc1
css, whitelabel: 0.48kb increase (0.04kb/gzip)
An insignificant difference was also found for: js, css (master), css (internal)

notlee (approved):

origamiserviceuser:
:tada: This PR is included in version v8.4.1 :tada:

PR #392, opened 2021-03-15
use pull_request_target and not pull_request for the safe workflows

JakeChampion:
We have workflows that currently fail to run correctly for pull-requests from forks or pull-requests from dependabot. The reason is because those workflows are being triggered by the 'pull_request' event, which does will use a restricted github api token for pull-requests from forks or dependabot for security reasons.

For our workflows which are 'safe' (meaning they do not checkout the code or execute the code in the repository), we can change their trigger from 'pull_request' to 'pull_request_target', which uses a github api token with more permissions, specifically it has permission to read/use secrets, which are the things are workflows require.

Our workflows are safe, they do not checkout or execute the code from the repository, so we should change to using `pull_request_target` to make our workflows work for dependabot and pull-requests from forks.

### DIFF
--- a/.github/workflows/add-new-issues-and-pull-requests-to-origami-project-board.yml
+++ b/.github/workflows/add-new-issues-and-pull-requests-to-origami-project-board.yml
@@ -2,7 +2,6 @@ name: Add new issues and pull requests to Origami Project Board
 on:
   - issues
   - pull_request_target
-  - pull_request
 
 jobs:
   origami_project_board_job:
--- a/.github/workflows/apply-labels.yml
+++ b/.github/workflows/apply-labels.yml
@@ -2,7 +2,6 @@ name: Apply Origami labels
 on:
   - issues
   - pull_request_target
-  - pull_request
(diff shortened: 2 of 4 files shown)


origamiserviceuser:
o-forms bundle size difference from 8.4.3 to abfe9c7131d861b415e695e7c8aa1f184e24d791

origamiserviceuser:
o-forms bundle size difference from 8.4.3 to 76ad5028128532e4a9d9911ec00470c0d7b75abb

origamiserviceuser:
o-forms bundle size difference from 8.4.3 to a08e022fbca9f0035e3e345b064d89356d715b1f

origamiserviceuser:
o-forms bundle size difference from 8.4.3 to 0169074a3976310ef23626ce27335f553b031a47

origamiserviceuser:
o-forms bundle size difference from 8.4.3 to 988d093672a8c7191168b40638cced10ddbcaa12

PR #398, opened 2021-05-07
We don't want to use dependabot to update dependencies anymore

JakeChampion:


origamiserviceuser:
o-forms bundle size difference from 8.5.1 to 311d3e94abfbd0dd509f2c020ada5076b1e868a5

PR #399, opened 2021-07-27
Update README/MIGRATION.md

notlee:
Based on user feedback/questions.

________

https://financialtimes.slack.com/archives/C02FU5ARJ/p1627311495242300?thread_ts=1627304115.241200&cid=C02FU5ARJ

Hey, I think I’ve figured out what’s happened here, these visual issues are due to changes we need to make for the v6->v7 migration.

If I’m not mistaken:
1. bigger (full width): v7 removed max-width from form fields because we found users often need to override that based on context. To shrink it back up either set a max-width on the field or place it within a containing element that has a max-width set.
2. spaced out: the label is set to a percentage width so that multiple fields with an inline label line up. As this field is really wide (because of 1) and the label text is quite short they appear spread out. You don’t have a form of multiple fields though so can add o-forms-title--shrink to the title element.
3. vertical alignment: it’s not vertically aligned by default to account for long labels / prompt information. Since you have a short label you can apply o-forms-title--vertical-center to the title element.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -69,7 +69,7 @@ The customisation mixin outputs a custom modifier, and can be applied as follows
 // the custom class modifier will be: .o-forms-input--my-theme
 ```
 
-The markup has been changed entirely to accomodate the following structure:
+The markup has been changed entirely to accommodate the following structure:
 ```
 ┌— field container (.o-forms-field) —————┐
 |      (one of div or label)             |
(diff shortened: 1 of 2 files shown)


JakeChampion (approved):

origamiserviceuser:
:tada: This PR is included in version v9.0.1 :tada:

